PR #117, opened 2025-01-15
skip caching in prettier job to speed up

haralde:


.github/workflows/prettier.yaml
```diff
@@ -1,30 +1,28 @@
...
+        strategy:
+            matrix:
+                node-version: [18]
+        steps:
+            - uses: actions/checkout@v3
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v3
+              with:
+                  node-version: ${{ matrix.node-version }}
+            - name: Install dependencies
```
kvashchuka, line 23: doesn't caching suppose to make it faster? or am I missing something?

haralde: It is the cleanup, compressing and saving of cache after the run that takes more than 2 minutes. No need to cache if the job it self takes 5 sec

haralde: And also we don't need to do a full npm install, we only need to install prettier to do the code formatting check

kvashchuka: Agree! I also think we should only check the files that were changed, so need to check everything in frontend folder. I think there was a flag for it, will do some googling

kvashchuka:
I checked Actions and prettier workflow on this PR is considerable faster than it was on previous PRs! Good change! Fetching changed files and only check them is not as straightforward as I thought it will be, so I think we can skip it for now.

PR #134, opened 2025-01-23
123 add date column to simulation and project list

SAKavli:
Solves #123


haralde (approved):
Looks good, but change date formatting as discussed

PR #137, opened 2025-01-24
Move backend api call to fronend api

SAKavli:
Most classes calls the front-end api to call the backend api. However, this was not the case for CreateProjectDialog, so I moved it to the front-end api to limit backend calls to the frontend api file.


haralde (approved):
Nice

PR #143, opened 2025-01-27
Disable button without projectId

SAKavli:
Solves #140


Hardins (approved):
Reviewed together with Sanders and Harald.

PR #163, opened 2025-02-10
Fix error when common_paths are empty

haralde:


frontend/src/pages/Reactions.tsx
```diff
@@ -44,7 +44,8 @@ const Reactions: React.FC<ResultsProps> = ({ simulationResults }) => {
             return <p>{chemicalFormula}</p>;
         }
     };
-
+    console.log(common_paths.paths);
```
Hardins, line 47: I just saw this, do we want to keep the console.logs for the main version?

PR #174, opened 2025-02-13
Code cleanup

SAKavli:
I imagine this is an example of how code can be better organized.
Pros:
- Components/functions are easier reusable
- Pages/components are shorter/more readable with imported sub-components/functions
- Better testability of components/functions
- Better developer experience with isolated components/functions

Cons:
- More files in repo
- Slightly more lines of code
- Harder to get overview of the repo


haralde (approved):
Looks good

PR #197, opened 2025-02-20
Add result scatterplot component

SAKavli:


lars-petter-hauge (approved):
Thanks!

Nice splitting of commits ;)

PR #201, opened 2025-02-24
Add custom legend to scatterplot

SAKavli:


frontend/src/functions/Formatting.tsx
```diff
@@ -1,5 +1,6 @@
 import { Data } from "plotly.js";
 import { SimulationResults } from "../dto/SimulationResults";
+import { IGraphData } from "../components/ResultScatterPlot";
```
lars-petter-hauge, line 3: Looks like we are getting a circular reference? Unsure how important it is, but It's a bit of a smell...

lars-petter-hauge:
Nice changes (and good commit structure!), this improves the plot by miles! I think it should go in and we can improve on it as we go ahead.

lars-petter-hauge:
Quick thought. I see a fair amount of these utility functions being added to separate files. I think that in overall, if a function is only used in a specific location it could be a (private - not exported) function within the place it is used. Adding them to separate files could give the impression that it is used more heavily, while if it is just in a single file and not exported we can rest assure that it is only used in one location (disregard that LSP will give you this information...).

Otherwise the utility files quickly grows to some kraken like thing, and can be demanding to maintain..

lars-petter-hauge (approved):
Approving even with some comments. Since it looks like the circular reference doesn't break things it's good. Check out how that works in React though

frontend/src/functions/Formatting.tsx
SAKavli: Agreed, also weird to have interface declaration and usage in the same file. I have moved into it's own file now!

SAKavli:
Regarding the utility files, the advantage is testing and less code to read in the components. But your point is valid, should be easily fixable if desired.

PR #206, opened 2025-02-27
Move getCurrentDate call to api

SAKavli:


frontend/src/pages/InputForm.tsx
```diff
@@ -65,7 +65,6 @@ const InputForm: React.FC = () => {
                     formConfig,
                     selectedModel,
                     simulationName,
-                    getCurrentDate()
```
kvashchuka, line 68: I believe you will have unused import in this file now

kvashchuka (approved):
This is an excellent example of an atomic change! Thank you @SAKavli 👍

PR #216, opened 2025-03-06
Edit experiment result type

SAKavli:
We have just made some changes which are waiting for approval from Oasis team. This change will change some naming conventions and remove unused items from the database.
This commit should accomodate for these changes, but need to be validated when the change is "merged" in Oasis.


SAKavli:
Reverted changes in Oasis database

Issue #239, opened 2025-03-21
Handle dates correctly, should be stored as ISO8601


haralde:
Seems that we are storing dates as strings in db.
To be able to search and filter on dates, they should be stored in ISO8601 format ("2024-03-20T00:00:00.000Z")

1. Update datamodels in backend with defaults?
2. If passing in dates from frontend, convert to 8601
3. Convert in frontend to a more userfriendly format when displaying

SAKavli:
Solved in #245

PR #245, opened 2025-03-24
Change date format

SAKavli:
Dates are now stored in ISO format in database and rendered in the same user-friendly way in the frontend.
Solves #239

### DIFF
--- a/frontend/src/api/api.tsx
+++ b/frontend/src/api/api.tsx
@@ -114,7 +114,6 @@ export const saveProject = async (name: string, description: string, isPrivate: 
             name,
             description,
             private: isPrivate,
-            date: new Date().toISOString(),
         }),
     });
     if (!response.ok) {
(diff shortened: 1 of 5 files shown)


frontend/src/api/api.tsx
```diff
@@ -216,7 +214,7 @@ export const saveSimulation = async (
             concs,
             settings,
         },
-        date: getCurrentDate(),
+        date: new Date().toISOString(),
```
haralde, line 217: Nice! Could also cosinder setting current date  as default on backend datamodel, instead of passing it from frontend:

class Scenario(BaseModel):
    id: UUID = uuid4()
    project_id: str = ""
    name: str = ""
    owner: str = ""
    scenario_inputs: SimulationRequest = SimulationRequest()
    model: str = "arcs"
    date: str = Field(default_factory=lambda: datetime.utcnow().isoformat())

haralde (approved):

PR #246, opened 2025-03-24
Fix stale simulation results

SAKavli:
We had the problem that simulation results didnt update when running a new simulation. This was because of the initial value in useQuery being cached. Instead of handling cached values, this edit gives us more control of which results to use and fixes the bug.


kvashchuka (approved):
Awesome! I was just talking to SM about it! Thanks for fixing!

PR #248, opened 2025-03-25
Clean up

SAKavli:


SAKavli:
image

SAKavli:
0 inputs gives this error. This is due to Arcs not returning all values given certain inputs because they aren't produced, which is not handled in the backend of Acidwatch. Last commit fix this.
image

Issue #204, opened 2025-02-25
Remove unnecessary <div>'s in return statements


SAKavli:
One can only return one parent component when returning components. 
When wanting to return sibling components together, a quick fix is to wrap it in a >div></div>. 
This is apparently a code smell. This creates unnecessarily complex elements in the browser, and can be fixed by replacing <div></div> with <React.Fragment></React.Fragment> or just <> </>.
However, this should be double checked to not cause any issues.

PR #249, opened 2025-03-25
Fix divs

SAKavli:
Solves #204


SAKavli:
For performance and code smell purposes, divs should not be
overused as they cause complexity for the DOM.
Also, generally, I believe components should describe it's
contents, while the styling may vary based on its use, which
should be decided one level up.
For example, project list needs a certain width to fit nicely
on the page. But this should be defined in the page where project
list is rendered, as this style could change for other pages
containing projectlist in the future.

Change my mind!

UsamaEquinorAFK (approved):
lgtm

PR #255, opened 2025-03-27
Improve conditional rendering

SAKavli:
As old results from api queries are cached, these should be used on render. Usequery will still query the api for new values,
 which are displayed should the value update.
In addition, this change will not stop the page from rendering if the query is either loading or having an error if the result is
 already cached.


UsamaEquinorAFK (approved):
lgtm

PR #273, opened 2025-04-29
Add continuous and bounded pressure and temperature inputs for arcs

UsamaEquinorAFK:


UsamaEquinorAFK:
There is some padding of white spaces, its just prettier making the work less pretty

haralde (approved):

PR #295, opened 2025-05-05
Friendly error message if no oasis access

### LINKED ISSUE #284: provide a good message to the users when they are missing access to the lab results

haralde:
fixes #284


frontend/src/api/api.tsx
```diff
@@ -361,8 +361,17 @@ export async function getLabResults(): Promise<ExperimentResult[]> {
             Authorization: "Bearer " + token,
         },
     });
+
     if (!response.ok) {
-        throw new Error("Network response was not ok");
+        if (response.status === 401) {
+            throw new Error("Unauthorized: Apply for access to CO2 lab results in AccessIT");
```
kvashchuka, line 367: Should it be "Apply ..." or "Log in to view experimental results"? Right now of course, you are either logged in from the beginning or you are not, so no way to test it in the app ...

haralde: This is 401 or 403 from Oasis, so you are already logged in to Acidwatch. Actually I'm not sure when Oasis sends 401, think it returns 401 if user does not have access to Oasis at all (for example external user), and 403 if user does not have access to the catalog. That is what they told me. But difficult to test before deployed to prod, since in Oasis test we all have full access:(

kvashchuka: Yeah, that's exactly what I was thinking - before the remove login to AcidWatch we will never reach 401, so no way to test as of now

kvashchuka: However, we can also merge it in and change if we see it's not working properly when we open AcidWatch

PR #308, opened 2025-05-14
Setting up devcontainer for opensource use

UsamaEquinorAFK:


UsamaEquinorAFK:
So far these steps are still manual
 
- Setup .env in backend
- Open ports to be public 
- Include backend api uri in cors exception for the arcs instance. (This can be a reference to localhost 8001 not codespace name)
- Get rid of codespace name incuded in backend api uri from frontend .env file

UsamaEquinorAFK:
Image 20-05-2025 at 15 01
Just adding this here to know that it worked at this point in time

lars-petter-hauge:
cool! Nice to iterate on this part :)

.devcontainer/install-dependencies.sh
```diff
@@ -0,0 +1,25 @@
...
+
+pip install poetry
+
+poetry install --directory ./backend
+
+deactivate 
+
+echo "VITE_API_URL=https://bookish-space-parakeet-6q5jv596x56c5r5w-8001.app.github.dev">>./frontend/.env
+
+npm install ./frontend
```
lars-petter-hauge, line 13: we should probably change directory before we do this.

As we saw we had to add the `node_modules` directory, even though it should already exist. The reason is that we expect it to be within the "frontend" directory.

this also goes for `package.json` and `package-lock.json`. We want to use the ones in the frontend, and not add another set.

(which also means we should remove those json files before merging)

.devcontainer/install-dependencies.sh
```diff
@@ -0,0 +1,25 @@
...
+
+source venv/bin/activate
+
+pip install poetry
+
+poetry install --directory ./backend
+
+deactivate 
+
+echo "VITE_API_URL=https://bookish-space-parakeet-6q5jv596x56c5r5w-8001.app.github.dev">>./frontend/.env
```
lars-petter-hauge, line 11: You mentioned this as an issue, I just a comment so we can mark as resolved once we know what to do

.gitignore
```diff
@@ -166,3 +166,4 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
 frontend/src/.DS_Store
+node_modules
```
lars-petter-hauge, line 169: once we fix changing directory we can remove this one as well

backend/src/acidwatch_api/app.py
```diff
@@ -26,7 +26,7 @@
 
 app.add_middleware(
     CORSMiddleware,
-    allow_origins=origins,
+    allow_origins=["*"],
```
lars-petter-hauge, line 29: we have to revert this before we merge. The codespace name is the most difficult one I reckon.. It really affects several parts of our work

frontend/src/main.tsx
```diff
@@ -4,31 +4,16 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
...
-                        console.log("Authentication error:", error);
-                        return <div>Error: {error ? error.message : "Unknown error"}</div>;
-                    }}
-                >
-                    <AppInsightsContext.Provider value={reactPlugin}>
-                        <App />
-                    </AppInsightsContext.Provider>
-                </MsalAuthenticationTemplate>
-            </QueryClientProvider>
-        </MsalProvider>
```
lars-petter-hauge, line 32: Yeah, we still need a way to authenticate in the case that the user would like to use a service that requires an access token. But we would like it to be optional instead, not a requirement. Unsure if we need some sort of "sign-in" button

lars-petter-hauge:
I suggest that we revert some of the changes (mostly what is commented), and include the rest (such as dockerfile and devcontainer), and get those things merged. It's okay that there are some parts missing, and also we can document things that  need to be manually done (and over time those that list will hopefully be reduced).

What do you think :)

UsamaEquinorAFK:
Too many comments I am sad

lars-petter-hauge:
> Too many comments I am sad

No no no, celebrate discussion! :)

ps: this pr is an order of magnitude off from my record: https://github.com/equinor/configsuite/pull/70 - and that PR I have reflected over multiple times, really enjoyed how we went about discussing first the testing/behaviour of we what we wanted before discussing implementation.

backend/src/acidwatch_api/app.py
UsamaEquinorAFK: Fixed now

frontend/src/main.tsx
UsamaEquinorAFK: Rebased to the new verison of main

frontend/src/api/api.tsx
```diff
@@ -83,13 +81,10 @@ export const runSimulation = async (formConfig: FormConfig, selectedApi: string)
 };
 
 export const getModels = async (): Promise<Record<string, ModelConfig>> => {
-    const token = await getAccessToken();
```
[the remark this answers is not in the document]
UsamaEquinorAFK: Fixed in new rebase i believe

docker-compose.yml
```diff
@@ -11,3 +11,11 @@ services:
...
     ports:
       - "8000:8000"
+    environment:
+      CONNECTION_STRING: "local"
+      CO2SPEC_API_BASE_URI: "http://localhost:5005"
+      ARCS_API_BASE_URI: "http://localhost:5001"
+  model_arcs:
+    build: https://github.com/equinor/arcs.git#test_compose_from_other_repo
+    ports:
+     - "5001:5001"
```
fmell, line 21: Could this be removed?

package-lock.json
fmell, line 1: This can be removed

package.json
fmell, line 1: This can be removed

package-lock.json
UsamaEquinorAFK: done.

package.json
UsamaEquinorAFK: done.

.devcontainer/install-dependencies.sh
UsamaEquinorAFK: done.

docker-compose.yml
UsamaEquinorAFK: done

haralde (approved):

PR #314, opened 2025-05-21
Use synthetic data on failure - prompt access

lars-petter-hauge:


lars-petter-hauge:
resolves [#285](https://github.com/equinor/acidwatch/issues/285)

PR #323, opened 2025-05-23
Prevent acidwatch from saving invalid simulation names

### LINKED ISSUE #322: Simulations can be named " " (space), meaning it's not possible to access it again
Image

Here I'm showing the HTML anchor tag in dev view. It's 0x0 size, meaning it's not possible to access.

UsamaEquinorAFK:
Closes #322

### DIFF
--- a/frontend/src/components/SaveResultButton.tsx
+++ b/frontend/src/components/SaveResultButton.tsx
@@ -94,6 +94,7 @@ const SaveResultButton: React.FC<{ props: SimulationProps }> = ({ props }) => {
         );
     }
 
+    const saveAble = !Boolean(selectedProjectId && simulationName && simulationName.trim()) && !isSimulationSaving;
     return (
         <>
             <Autocomplete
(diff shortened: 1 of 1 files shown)


UsamaEquinorAFK:
[
Image 23-05-2025 at 14 43
Image 23-05-2025 at 14 42
](url)

UsamaEquinorAFK:
> Cheers,
> 
> Please note, after asking for a review, please only push commits (not force push). It can become difficult to review when the "REFRESH" button start popping up.. what happens when I refresh, how do I know what has changed?
> 
> And then in the end we fix commits, squashing what needs squashing :)
> 
> (a nice way is to make `--fixup` commits, which can be nicely handled afterwards)

Yes I understand but I only did prettier write . :D

frontend/src/components/SaveResultButton.tsx
```diff
@@ -124,13 +124,15 @@ const SaveResultButton: React.FC<{ props: SimulationProps }> = ({ props }) => {
             ) : (
                 <Button
                     onClick={handleSave}
-                    disabled={!Boolean(selectedProjectId && simulationName) && !isSimulationSaving}
+                    disabled={
+                        !Boolean(selectedProjectId && simulationName && simulationName.trim()) && !isSimulationSaving
+                    }
```
lars-petter-hauge, line 129: Although I think the entire 122 to 137 could have been refactored out of this function (as it is nested within the return statements, doesn't look very nice..)

A suggestion though, how about setting for instance:

```ts
const saveAble = !Boolean(selectedProjectId && simulationName && simulationName.trim()) && 
!isSimulationSaving;
return (
....
disabled={saveAble}
```
to reduce code complexity. Perhaps also have a comment as to what why we validate it like this, if you think it makes sense

lars-petter-hauge (approved):

PR #327, opened 2025-05-26
Introduce conditional rendering of simulations table

### LINKED ISSUE #318: Empty project page should have the text "This project is empty" somewhere
Image

Currently it looks like it's bugged.

UsamaEquinorAFK:
Fixes #318

### DIFF
--- a/frontend/src/pages/SimulationList.tsx
+++ b/frontend/src/pages/SimulationList.tsx
@@ -92,63 +92,69 @@ export default function SimulationList(): JSX.Element {
             </StyledRowLayout>
             <br />
             <StyledRowLayout>
-                <Table style={{ width: "100%" }}>
-                    <Table.Head>
-                        <Table.Row>
-                            <Table.Cell>Name</Table.Cell>
-                            <Table.Cell>Created by</Table.Cell>
-                            <Table.Cell>Creation date</Table.Cell>
-                            {isProjectYours && <Table.Cell>Actions</Table.Cell>}
-                        </Table.Row>
-                    </Table.Head>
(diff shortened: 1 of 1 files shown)


frontend/src/pages/SimulationList.tsx
```diff
@@ -92,63 +92,69 @@ export default function SimulationList(): JSX.Element {
...
+                                            </Table.Cell>
+                                        )}
+                                    </Table.Row>
+                                );
+                            })}
+                        </Table.Body>
+                    </Table>
+                ) : (
+                    <Typography>No simulations available. The project is currently empty.</Typography>
+                )}
```
lars-petter-hauge, line 157: A quick suggestion is to do this the other way, checking if   (pardon syntax errors) 
```ts
!simulations || simulations.length == 0 ? 
return <Typography>No simulations available. The project is currently empty.</Typography> :
...
```

We are already in the return here, so unsure if this is a good way of dealing with it. Still improving javascript ways of working, so not sure how I would do it. But surely having these nested conditionals, with the "else" all the way in the end tend to be less readable, than to check in the top and early return for the simple things.

lars-petter-hauge: I see there are multiple nested here in this section, could be improved here as well (for this PR though I suggest only to do the checking up top and we leave the refactoring for later when we have a tad bit more time)

UsamaEquinorAFK: Done.

lars-petter-hauge (approved):
Thanks!

PR #330, opened 2025-05-26
Use correct co2 spec keys for results

### LINKED ISSUE #325: Change the graph in co2spec to show change in concentration and not fin al conc

lars-petter-hauge:
Closes #325


lars-petter-hauge:
Updated branch to maim

PR #336, opened 2025-05-27
Fix input concentration max ppm to 1000000

### LINKED ISSUE #329: ppm cannot be larger than 1 million (by definition)

UsamaEquinorAFK:
Closes #329 329

### DIFF
--- a/backend/src/acidwatch_api/models/model_config.py
+++ b/backend/src/acidwatch_api/models/model_config.py
@@ -7,6 +7,7 @@ def get_model_config():
                     "H2O": {
                         "defaultvalue": 30,
                         "meta": "ppm",
+                        "max": 1000000,
                         "type": "float",
                         "input_type": "textbox",
                         "enabled": True,
@@ -14,6 +15,7 @@ def get_model_config():
                     "O2": {
                         "defaultvalue": 30,
                         "meta": "ppm",
+                        "max": 1000000,
                         "type": "float",
                         "input_type": "textbox",
                         "enabled": True,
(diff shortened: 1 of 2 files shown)


UsamaEquinorAFK:
Image 27-05-2025 at 12 55

backend/src/acidwatch_api/models/model_config.py
```diff
@@ -76,160 +81,183 @@ def get_model_config():
                     "CH2O2": {
                         "defaultvalue": 0,
                         "meta": "ppm",
+                        "max": 1000000,
                         "type": "float",
                         "input_type": "textbox",
                         "enabled": False,
                     },
```
lars-petter-hauge, line 88: close to 200 lines of copy paste here..

Could be something along:

```python
{
    comp: {
        "defaultvalue": 0,
        "meta": "ppm",
        "max": 1000000,
        "type": "float",
        "input_type": "textbox",
        "enabled": False,
    }
    for comp in ["CH202", ...]
}
```

What do you think?

UsamaEquinorAFK: 20*
But yes I agree

lars-petter-hauge: I see some variations in default value, those could be separated:

```python
DEFAULT_VALUES = {"H2O": 0...}

{
    comp: {
        "defaultvalue": DEFAULT_VALUES.get(comp, 0),
        "meta": "ppm",
        "max": 1000000,
        "type": "float",
        "input_type": "textbox",
        "enabled": False,
    }
    for comp in ["CH202", ...]
}
```

UsamaEquinorAFK: Ah now I understand yes it can be 200 :D

lars-petter-hauge:
Nice work getting these things fixed :)

pinkwah (approved):

Issue #335, opened 2025-05-27
Moving between models should show respective results


lars-petter-hauge:
Currently, running a calculation on the arcs model and then moving to co2spec will keep the results from the arcs run in the co2spec tab.

https://github.com/user-attachments/assets/0576a561-09d3-4964-9a66-0a716349c392

PR #343, opened 2025-05-28
Clear simulation results upon switching models

UsamaEquinorAFK:


UsamaEquinorAFK:
Closes #335

UsamaEquinorAFK:
> This seems a bit heavy-handed. If the user spent 5 minutes running an ARCS thing then accidentally switch to CO2Spec, will this delete their ARCS results?

No it wont, it will keep the results

UsamaEquinorAFK:
Will be refactored

PR #345, opened 2025-06-02
Remove authentication from acidwatch for running open source models

UsamaEquinorAFK:


frontend/src/pages/InputForm.tsx
```diff
@@ -174,7 +174,8 @@ const InputForm: React.FC = () => {
             <div style={{ marginLeft: "100px" }}>
                 {isSimulationRunning && <img src={loader} alt="Loading" style={{ width: "70px" }} />}
                 {simulationResults && (
-                    <><h3>Save this simulation?</h3>
```
lars-petter-hauge: I remember we did something adding unique key, but I can't say that it is part of this commit? This seems to only include a line change?

PR #346, opened 2025-06-03
fix: project access, user is able to se public projects results

### LINKED ISSUE #340: Bug: "Failed to fetch" when trying to view results from internal projects

haralde:
fixes #340


UsamaEquinorAFK (approved):
lgtm

PR #360, opened 2025-06-06
Tools for oasis upload and delete, and run arcs from experiments

haralde:
Some scripts to add data to oasis and acidwatch. Might need some tweaking to work.


kvashchuka:
Let's put it in Wiki

PR #364, opened 2025-06-12
Add Architecture Decision Record

lars-petter-hauge:


fmell:
I think this is a great starting point! Maybe we could try to work on including a bit more of the reasoning behind our decision?

Also.. 

- Should we find a good ADR template to use? I think there's even some standardised ways to do markdown ADRs where theres some tooling available. (https://adr.github.io/adr-tooling/)
- I am missing the status section where it says if its proposed, accepted, rejected, superseded and so on.
- I also think the ADRs should be stored numbered and with a descriptive filename so that they appear in the correct order

haralde (commented):
I think this looks good! Maybe include the templeate you used as well.

PR #369, opened 2025-06-16
Remove unused arcs API scope

fmell:


lars-petter-hauge (approved):
Arcs no longer requires auth, good to go!

PR #373, opened 2025-06-18
Adapter

pinkwah:


backend/src/acidwatch_api/models/example_adapter.py
```diff
@@ -0,0 +1,77 @@
...
+    #
+    # This function must return a dict that maps a valid chemical substance
+    # (eg. 'HNO2', but not 'ASDF') to a floating point value between 0 and
+    # 1,000,000 (in ppm) or None, indicating it's possible for the user to
+    # supply it, but it's disabled by default.
+    #
+    # The exception is CO2, which must not be specified as it's the solvent.
+    concs = {
+        "HNO": 10,
+    }
```
lars-petter-hauge, line 57: I kinda dislike that each model needs their own default value.. Does it not make more sense for the models to simply determine:

`accepted_components = ["HNO", "H2O"]`

And AcidWatch sets a default value for the most familiar ones, and 0 for the rest? Reduces complexity a bit for this, and might make it simpler for the user when they switch between models (as in, all models have same defaults).

PR #374, opened 2025-06-18
Change from CO2spec to ToCoMo

lars-petter-hauge:
Only user phasing change. Do want to change all other locations


pinkwah:
Is this a breaking change in the database?

lars-petter-hauge:
> Is this a breaking change in the database?

We never use the model name in the frontend, so it won't matter per say.

Is it ugly? Yes.

lars-petter-hauge:
So I went for the very simple, technical debt, kind of entry..

pinkwah (approved):

PR #393, opened 2025-07-28
Move save result control and disable after save

haralde:
Disable the button after it is pressed and also moved the inputs to keep the results tabs fixed at top. Open for suggestions, it could also be a popup, but I find the current location above the results tab taking to much focus away from the results

image


kvashchuka:
Pop up can be cool, but I don't think we should spend time implementing popup when we can just save everything - in my opinion this is the expectation from a browser app nowadays. Maybe popup saying that "Oh, you are not logged in - log in to save simulations"?

PR #394, opened 2025-07-28
[WIP] start refactor

pinkwah:


lars-petter-hauge:
Although not directly part of this PR, we change som SimulationResult to RunResponse here, and I think we included too much in the RunResponse. Let's remove `model_config`, we never return that in SimulationResult - it shouldn't be in RunResponse.

```
class RunResponse(BaseModel):
    model_config = ConfigDict(
        alias_generator=to_camel,
        populate_by_name=True,
        from_attributes=True,
    )

    initial_concentrations: dict[str, int | float]
    final_concentrations: dict[str, int | float]
    visualizations: Iterable[AnyResult] = ()

```

frontend/src/api/api.tsx
```diff
@@ -62,8 +31,8 @@ export const runSimulation = async (formConfig: FormConfig, selectedApi: string)
                 Authorization: `Bearer ${token}`,
             },
             body: JSON.stringify({
-                concs: absoluteConcentrations,
-                settings: settings,
+                concs: concentrations,
+                settings: parameters,
```
lars-petter-hauge, line 35: could we not rename this as well? We are naming them `parameters` in the front-end. And also `parameters` in the adapters. It's technically only in the interface of the backend api that we are naming it `settings`. Let's do ourselves a favour and rename to `parameters` in the web api (so `RunRequest`)

frontend/src/components/ConcResultTable.tsx
```diff
@@ -1,20 +1,10 @@
...
-            [key: string]: number;
-        };
-        final: {
-            [key: string]: number;
-        };
-        change: {
-            [key: string]: number;
-        };
-    };
-};
```
lars-petter-hauge, line 15: anyone have any thoughts of this part? I think I made the change to not use `initFinalDiff`, as the change is always implicit, and it adds another layer. At the same time we need to recalculate the diff relatively often..

Personally I prefer to keep the change (using `initialConcentrations` etc), but if everyone else is opposed I won't have (too) strong opinions.. :D

frontend/src/components/ModelInputs.tsx
```diff
@@ -0,0 +1,127 @@
...
+                }}
+                disabled={selected === null}
+            >
+                +
+            </Button>
+        </div>
+    );
+}
+
+function ParametersInput({
```
lars-petter-hauge, line 43: I think this is more general than "ParametersInput". Technically this could be together with `ListWithItemAdder`, just without any items to add.. Would be nice not having to specialize these components..

frontend/src/dto/SimulationResults.tsx
```diff
@@ -1,27 +1,14 @@
...
-            frequency: Record<string, number>;
-        };
-    };
-    chart_data: {
-        comps: { [key: string]: string };
-        values: { [key: string]: number };
-        variance: { [key: string]: number };
-        variance_minus: { [key: string]: number };
-    };
+    modelConfig: ModelConfig;
```
lars-petter-hauge, line 10: Suggested to remove the `ModelConfig` from `RunResponse`, we might want to remove it from `SimulationResults` as well?

Although, we do want to know the name of the model.. And I would argue even more, like model version.. 

Likewise with suggested on the other sides, rename `visualizations` to something else.

frontend/src/pages/Results.tsx
```diff
@@ -48,23 +27,18 @@ const Results: React.FC<ResultsProps> = ({ simulationResults }) => {
...
-                                    ? simulationResults.results.initfinaldiff
-                                    : fetchedResults!.results.initfinaldiff
-                            }
+                            initialConcentrations={simulationResults.initialConcentrations}
+                            finalConcentrations={simulationResults.finalConcentrations}
                         />
                     </Tabs.Panel>
                     <Tabs.Panel>
-                        <Reactions simulationResults={simulationResults ? simulationResults : fetchedResults!} />
+                        {/* <Reactions simulationResults={simulationResults.initial_concentrations, simulationResults.final_concentrations} /> */}
```
lars-petter-hauge, line 37: I think we can remove this reactions panel altogether in this PR. We want to make this Panel generation more general, and depending on the number of those `JSON | test` entries that is included..

backend/src/acidwatch_api/app.py
```diff
@@ -128,11 +126,19 @@ async def run_model(
...
-    return _legacy.result_to_simulation_results(request.concs, result)
+    runid = uuid4()
+    await _run_adapter(adapter, runid)
+    result = RESULTS[runid]
+
+    if isinstance(result, BaseException):
+        raise HTTPException(
+            status_code=400,
+            detail=f"Model failed to calculate the change: {format_exception(result)}",
+        )
```
lars-petter-hauge, line 140: I made the status code 400, I think it's probably incorrect.. It might not be bad request just because the model fails. I wondered about 500 as well, but it's not really internal error, is it?

lars-petter-hauge: A bit further down we do:

``` 
   if isinstance(result, BaseException):
        return {"error": format_exception(result)}
```

We probably want to be consistent in how we handle errors in our api. We could change to this or do fixes afterwards - we don't need to do everything now.

frontend/src/components/ModelInputs.tsx
```diff
@@ -0,0 +1,127 @@
...
+}> = ({ model, visible: not_hidden, onSubmit }) => {
+    const [concentrations, setConcentrations] = useState<Record<string, number>>(DEFAULTS);
+    const [visible, setVisible] = useState<string[]>(Object.keys(DEFAULTS));
+    const [parameters, setParameters] = useState<Record<string, number>>(getParameterDefaults(model));
+
+    if (!not_hidden) {
+        return <></>;
+    }
+
+    const invisible = model.validSubstances.filter((subst) => !visible.includes(subst));
```
lars-petter-hauge, line 91: this hidden vs invisible.. Let's rename this to `addableItems` or `itemsToAdd` or something similar.

UsamaEquinorAFK (approved):
will take care of requested changes later.

PR #396, opened 2025-07-30
feat: add solubility model

Sviatose:


lars-petter-hauge:
First off, really cool to see that you added a model using the adapter feature. The part in backend/src/acidwatch_api/models/solubilityccs.py looks pretty much what is the intention!

Did you notice the things you needed to do in the frontend though? The checking if different results exist etc? And when adding something in the solubility it was necessary to do something like 

```
// Only show the table for solubilityccs (when table_data is present)
const isSolubilityCCS = Boolean(currentResults.table_data);
```

inside the very general Results page.

What we want to do is to make the frontend more oblivious to the models. And also avoid those transformations in the backend (found in `_legacy.py`) - just leftover from the adapter work only beeing applied to backend. This is pretty much what the following PR does: https://github.com/equinor/acidwatch/pull/394. Currently, this PR goes a bit in the other direction (which technically is on par with what we already have), by adding model specific knowledge in the frontend. We would like that part to be more intuitive and scale better.

We do know that models provide different results, and those are not always easy to generalise. We are likely going to need some sort of model specific components in acidwatch (for instance the `common_paths` for arcs). Those will however be treated in a similar fashion as the adapters in the backend. The less "special treatment" a model needs, the easier it will be to add it.

Noticed that you added table data, agree that this part is necessary to add. We would like this to be an option to all models though.

Most of the things in this PR should be possible to handle, however we do not currently have a good way of communicating cross validation in the frontend, such as this:
```
 // Only apply disabling logic for solubilityccs
  let disabled = false;
  if (selectedModel === "solubilityccs") {
      if (
          key === "H2SO4" &&
          formConfig.inputConcentrations["HNO3"]?.defaultvalue > 0
      ) {
          disabled = true;
      }
      if (
          key === "HNO3" &&
          formConfig.inputConcentrations["H2SO4"]?.defaultvalue > 0
      ) {
          disabled = true;
      }
```
Cross validation might need to be added later, and for now have the adapter in the backend return a validation error (meaning the user will not see boxes changing colour, but running the simulation will fail with a good explanation). At least in the first iteration.

We will work on getting https://github.com/equinor/acidwatch/pull/394 merged, and then help out with this PR :)

lars-petter-hauge (approved):
Let's get this in and deal with the minor issues afterwards

PR #401, opened 2025-08-05
Use pyproject entry-points to discover models

pinkwah:
Instead of directly importing models, we can use Python's entry-points structure to dynamically find models when they're installed. This means that adding a new compatible model to AcidWatch requires the user to pip-install a package that defines a `acidwatch-model` entry-point.


pinkwah:
Good and interesting WIP @pinkwah! But closing because of lack of activity.

PR #406, opened 2025-08-06
use valid_substances for defaults

lars-petter-hauge:
Not only for the list items is it necessary, but also for the
concentrations. Otherwise the default items will be included whenever we
call the adapter backend.


lars-petter-hauge:
As seen in screenshot: solubility ccs used to receive all default values:

Screenshot 2025-08-06 at 14 24 05

kvashchuka (approved):

PR #408, opened 2025-08-06
Add temperature toggle (K / °C) to frontend

pinkwah:


frontend/src/contexts/SettingsContext.tsx
```diff
@@ -0,0 +1,90 @@
...
+    name = "kelvin";
+    unit = "K";
+}
+
+class Celsius extends Unit {
+    name = "celsius";
+    unit = "°C";
+
+    valueToNumber(value: number): number {
+        return value - 273;
```
kvashchuka, line 29: We might want to do 273.15

backend/src/acidwatch_api/models/solubilityccs.py
```diff
@@ -15,14 +15,14 @@ class SolubilityCCSParameters(BaseParameters):
     temperature: float = Parameter(
         15,
         label="Temperature",
-        unit=Unit.CELCIUS,
+        unit=Unit.TEMPERATURE_KELVIN,
         min=-100,
```
kvashchuka, line 19: For Kelvin temperatures cannot be below 0, so make sense to have min 0 and max 400

PR #409, opened 2025-08-06
Add gif for loading results

UsamaEquinorAFK:


UsamaEquinorAFK:
Zohar suggested we should go back to using tanstack usequery since it gives isLoading, should we go back ? Why did we remove it in the first place ? @lars-petter-hauge

kvashchuka:
I tried running it locally, and it doesn't work for ARCS, but works for solubility model. Do you see the same @UsamaEquinorAFK ?

PR #419, opened 2025-08-15
Add model endpoint tests

pinkwah:


backend/tests/test_models_endpoints.py
```diff
@@ -1,24 +1,216 @@
...
+    assert response.json() == {
+        "initialConcentrations": {},
+        "finalConcentrations": {},
+        "panels": [],
+    }
+
+
+@pytest.mark.parametrize(
+    "valid_substances,concs,expected_concs",
+    [(["H2"], {"H2": 10}, None), ([], {"H2": 10}, {}), (["H2"], {}, {"H2": 0.0})],
```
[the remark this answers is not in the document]
lars-petter-hauge: `test_dummy_model_only_valid_substances_are_present` is not correct, as in one of the places we are providing `H2` although we state that there are not valid substances.

backend/tests/test_models_endpoints.py
```diff
@@ -1,24 +1,216 @@
...
+
+    response = client.get_json("/models")
+
+    assert response == [
+        {
+            "accessError": None,
+            "displayName": "Dummy Model",
+            "modelId": "dummy",
+            "parameters": {
+                # Notice that it's "someField" and not "some_field"
```
lars-petter-hauge, line 128: this is because of the `toCamelCase` thingy in pydantic?

backend/tests/test_models_endpoints.py
```diff
@@ -1,24 +1,216 @@
...
+            {"someField": 123},
+            {"someField": 123},
+            id="valid camel case param",
+        ),
+        pytest.param(_DummyModelParameters, {}, {"someField": 0}, id="default value"),
+        pytest.param(
+            _DummyModelParameters,
+            {"some_field": 123},
+            {"someField": 123},
+            id="valid snake case param",
```
lars-petter-hauge, line 166: so we pretty much just handle both ways of stating parameter names in our api?

backend/tests/test_models_endpoints.py
```diff
@@ -1,24 +1,216 @@
...
+        ),
+        pytest.param(
+            _DummyModelParametersMinMax,
+            {"someField": 100},
+            "Input should be less than or equal to 5",
+            id="max",
+        ),
+    ],
+)
+def test_dummy_model_only_valid_parameters_are_present(
```
lars-petter-hauge, line 188: should we also test invalid parameters?

backend/tests/test_models_endpoints.py
```diff
@@ -1,24 +1,216 @@
...
+        pytest.param(
+            _DummyModelParameters,
+            {"some_field": 123},
+            {"someField": 123},
+            id="valid snake case param",
+        ),
+        pytest.param(
+            _DummyModelParameters,
+            {"someOtherField": 123},
+            "Extra inputs are not permitted",
```
lars-petter-hauge, line 171: Test says "test_dummy_model_only_valid_parameters_are_present", but in this case we have invalid parameters I suppose..?

I think this is nice to include in the test.

I wonder though, I think the error message should include the name of the invalid parameter.

backend/tests/test_models_endpoints.py
pinkwah: Yeah apparently Pydantic does both. I figure there's no point in changing it, but it's good to have tests that check in case the behaviour changes. Note that the frontend will always use the camelCase version, but someone using the API directly might end up using snake_case.

backend/tests/test_models_endpoints.py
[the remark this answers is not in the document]
pinkwah: In this test, if `expected_concs` is `None`, then we expect `concs` instead. But it's so confusing that I was scratching my head as well. We expect to have 10ppm H2 before and after.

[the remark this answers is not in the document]
lars-petter-hauge: `concs if expected_concs is None else expected_concs,`

I see.. I would really just have them parametrize use `(["H2"], {"H2": 10},  {"H2": 10})`. Much easier to read

backend/tests/test_models_endpoints.py
lars-petter-hauge: Hopefully this isn't something that could potentially bite us. I think we should be good.. "famous last words"

backend/tests/test_models_endpoints.py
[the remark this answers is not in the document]
lars-petter-hauge: Could we also consider behaviour of: `([], {"H2": 10}, {})`, which I personally don't particularly like.

For instance, I suppose then that the following would be  possible: `(["H2O"], {"H2": 10, "H2O":5}, {"H2O":5})`. Now how would one interpret the results? Does it mean "H2" all of a sudden has vanished? Should the caller have business logic to say "well, if the substance isn't in the output, then it should be the same.."? Does it mean it doesn't react? We know of course that it is just ignored..

I would argue this is a fail, `ValueError` and raised to `HTTP_422_UNPROCESSABLE_ENTITY`.

What do you think?

PR #433, opened 2025-08-20
feat: Gibbs Minimization Model 

Sviatose:


lars-petter-hauge:
Cool, looks like less work than last time!

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -0,0 +1,155 @@
...
+    
+    model_id = "gibbs_minimization"
+    display_name = "Gibbs Minimization Model"
+    parameters: GibbsMinimizationModelParameters
+    base_url = configuration.ARCS_API_BASE_URI  
+
+
+    async def run(self) -> RunResult:
+        # Impurities in CO2
+        from neqsim import jneqsim
```
lars-petter-hauge, line 68: Imports are usually placed on top of the file

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -0,0 +1,155 @@
...
+            results[formula_name or neqsim_name] = mole_fraction
+
+        # Return results in expected format
+        # Return as tuple (final_concentrations, ReactionPathsResult) for RunResponse
+        return (
+            dict(results),
+            ReactionPathsResult(
+                common_paths={"paths": {}, "k": {}, "frequency": {}},
+                stats={"index": {}, "k": {}, "frequency": {}}
+            ),
```
lars-petter-hauge, line 154: can be removed, assuming this comes from arcs

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -115,7 +115,7 @@ async def run(self) -> RunResult:
         reactor = jneqsim.process.equipment.reactor.GibbsReactor("Gibbs Reactor", inlet_stream)
         reactor.setUseAllDatabaseSpecies(False)
         reactor.setDampingComposition(0.01)
-        reactor.setMaxIterations(10000)
+        reactor.setMaxIterations(20000)
```
lars-petter-hauge, line 118: What is the approximate time to complete on this calculation?

backend/src/acidwatch_api/models/gibbs_minimization_model.py
lars-petter-hauge: Then there's likely an issue with how `results` is returned instead. I can check

backend/src/acidwatch_api/models/gibbs_minimization_model.py
Sviatose: Thats a good question. It depends on the impurities and type of Equation of State chosen, there is no "time" stop, only iteration stop

Sviatose: I can make it run in a thread so we will have control over time

backend/src/acidwatch_api/models/gibbs_minimization_model.py
lars-petter-hauge: Uhm, but this adapter does not use anything from solubilityccs package? how does this work? Sounds like importing solubilityccs will modify the jneqsim package then?

lars-petter-hauge: (just tested returning `result` directly and did not encounter any issues.. feel free to try again and post any error)

backend/src/acidwatch_api/models/gibbs_minimization_model.py
Sviatose: Yes, so I suggest to include it so we are on a safe side of the time calculation so it will not run forever - I made a commit :)

Sviatose:
> image

This error is just annoying, not sure why we get it printed - it just says that some of the components are 0

backend/src/acidwatch_api/models/gibbs_minimization_model.py
lars-petter-hauge:
```
        thread = operations.runAsThread()
        thread.join(max_time_calculation_min*60*1000) #milliseconds 
        if thread.isAlive():
            thread.stop()
            raise CalculationTimeout(f"The calculation did not finish in time limit ({max_time_calculation_min} minutes).")
```

Please wait with this, we want to handle how adapters utilize threads and cores in a similar way, and at least ensure that it is the fastapi (uvicorn) that is in charge of it.

lars-petter-hauge:
> This error is just annoying, not sure why we get it printed - it just says that some of the components are 0

I am not getting this error? What input do you use to get this?

backend/src/acidwatch_api/models/gibbs_minimization_model.py
lars-petter-hauge: >  And now I see that the model uses the database defined in solubility package and not in neqsim

Sorry, but how do you see that?

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -0,0 +1,155 @@
...
+        reactor.setEnergyMode(jneqsim.process.equipment.reactor.GibbsReactor.EnergyMode.ISOTHERMAL)
+        reactor.run()  
+
+        assert inlet_stream.getFluid().getNumberOfPhases() == 1, "Gibbs model cannot work with two phases as of now" # Would be nice to show to the user 
+
+        # Get the outlet system
+        outlet_system = reactor.getOutletStream().getThermoSystem()
+
+        # Check mass balance convergence
+        assert reactor.getMassBalanceConverged(), "Mass balance should be converged" # Would be nice to show to the user
```
lars-petter-hauge, line 133: same with this as above, could you provide an example where this will trigger?

lars-petter-hauge:
I had some final comments on this. I think we afterwards can merge this (do a squash and merge).

Could you first run `poetry run ruff format` though?

backend/src/acidwatch_api/models/gibbs_minimization_model.py
Sviatose: I cannot provide example, but I has experienced it was triggered when there were bugs in code. I hope there are no bugs now, but we never know. Thats just for safety

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -0,0 +1,155 @@
...
+        # Create a Gibbs reactor
+        reactor = jneqsim.process.equipment.reactor.GibbsReactor("Gibbs Reactor", inlet_stream)
+        reactor.setUseAllDatabaseSpecies(False)
+        reactor.setDampingComposition(0.1)
+        reactor.setMaxIterations(20000)
+        reactor.setConvergenceTolerance(1e-3)
+        reactor.setEnergyMode(jneqsim.process.equipment.reactor.GibbsReactor.EnergyMode.ISOTHERMAL)
+        reactor.run()  
+
+        assert inlet_stream.getFluid().getNumberOfPhases() == 1, "Gibbs model cannot work with two phases as of now" # Would be nice to show to the user
```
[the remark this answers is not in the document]
Sviatose: Oh thats a good point, I moved it down because of work with threads, but if we dont have threading, then we just put it above the reactor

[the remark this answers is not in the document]
Sviatose: We will get this issue, for example, when you put a lot of water (>5000 ppm). The reactor designed to work only with one phase - CO2 - liquid water will break reactor model. Its possible to add "ideal gas law" maybe, that I will add, then we dont have this issue if we assume ideal gas.

Sviatose:
> > This error is just annoying, not sure why we get it printed - it just says that some of the components are 0
> 
> I am not getting this error? What input do you use to get this?

I think its a warning actually ... but I used all 10 default impurities and 100 bar and 25C, that is in the backend log

backend/src/acidwatch_api/models/gibbs_minimization_model.py
Sviatose: Because you get issue of "not present component database" ... that will be only with downgraded version of solubilityCCS that have its own database that is put in neqsim. So when I saw that neqsim was the latest version and does have all components, I had to update the solubilityCCS database as well (we dont need all components in solubilityCCS)

lars-petter-hauge:
Ref description, we just merged being able to provide adapter description: https://github.com/equinor/acidwatch/blob/c0f5696e05b8ace45146313f118d184ea591e3c1/backend/src/acidwatch_api/models/datamodel.py#L43

Feel free to add the disclaimer there :)

lars-petter-hauge:
I rebased on main, ran ruff, added description and pushed it.

PR #435, opened 2025-08-20
Use the same labResults for populating the plot

lars-petter-hauge:
Instead of trying to get the data out of the eds component row, we use
the initial data structure. Also rewriting slightly the transformation
function into scatter graph allows us to optionally filter on those
components.


haralde (approved):
nice!

PR #442, opened 2025-08-25
Use dashes to signify zeroes in the conc table

pinkwah:
Makes it a tad bit more readable. The value needs to be exactly zero, hence the `0.000` which seems out of place, but that means that it's just a small non-zero quantity.

image


lars-petter-hauge (approved):
Hopefully @UsamaEquinorAFK got his question answered :) 

If I could add any, it does seem a bit odd that input and output are both 0, and still there's a change. I understand the reason, but the user might be a tad confused.

PR #444, opened 2025-08-26
Add devenv integration

pinkwah:
I use Nix & devenv to handle most of my dependencies. I don't expect anyone else to upkeep this file, but it'd be nice to have it


UsamaEquinorAFK (approved):
approved to make Zohars life easier

PR #457, opened 2025-08-29
Include input details in result page

haralde:
Previously, when clicking a simulation in the list, you where sent to result page without any info on settings used for the simulation. This adds a card with relevent info:

image

### DIFF
--- a/frontend/src/api/api.tsx
+++ b/frontend/src/api/api.tsx
@@ -128,6 +128,22 @@ export const deleteProject = async (projectId: string) => {
         console.error("Error deleting project:", error);
     }
 };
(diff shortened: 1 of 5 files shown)


frontend/src/api/api.tsx
```diff
@@ -128,6 +128,22 @@ export const deleteProject = async (projectId: string) => {
         console.error("Error deleting project:", error);
     }
 };
+export const getSimulation = async (projectId: string, scenarioId: string): Promise<Simulation | null> => {
```
lars-petter-hauge, line 131: Thought we had this already? or something very similar

PR #472, opened 2025-09-08
[RFC] Port database to relational database

pinkwah:
This uses SQLAlchemy and uses basic FastAPI and Pydantic features to make everything work.

I would like us to switch from CosmosDB to a relational database for the following reasons:
1. We should stick to free and open-source software as much as possible
2. We shouldn't vendor lock ourselves with no visible upside
3. This code supports both SQLite and PostgreSQL for development and production, respectively. We can therefore get rid of the two clients.
4. FOSS means that it's very easy to have the actual production database software locally or on GitHub Actions. (Ie, easy to set up PostgreSQL locally)
5. I don't like the Facade pattern: Having a big class with a bunch of utilities isn't good code (`DBClient`)
6. Getting rid of the separated clients means we can make more effective use of FastAPI's dependencies concept. I have done this in this PR with `get_project` and `get_owned_project`.
7. Using SQLAlchemy, the code itself describes the schema of the database. With CosmosDB (or other NoSQL) this isn't the case, and you need to interact with the database directly.
8. SQL will check invariants for you and protect the data. With little code we get protection from foreign keys pointing to nowhere, or setting values to `NULL` that shouldn't be.
9. SQLAlchemy's Alembic makes migrations safe. If a transaction fails, the whole migration will fail and we'll be back in a safe state. The aforementioned protections afforded by the database are available.
10. Both SQLite and PostgreSQL support JSON data. They both allow querying this data as well. (Eg, fetching all simulations where SO<sub>2</sub> is greater than 5ppm is possible even if it's in a JSON column. But I'm doubtful if we'll ever need to do this anyway.)

If we're especially allergic to relational databases, I would propose to switch to MongoDB for the FOSS & vendor lock-in reasons mentioned above. But I will always argue that it's better to describe the scheme rather than dump everything into a database. Most data is very simple and relational, so we should make sure that we describe as much of it as possible, as early as possible.

### DIFF
--- a/backend/model.py
+++ b/backend/model.py
@@ -2,7 +2,18 @@ from __future__ import annotations
 from datetime import datetime
 from typing import Any
 from uuid import UUID, uuid4
-from sqlalchemy import ARRAY, JSON, DateTime, Float, ForeignKey, String, Uuid, create_engine, select, func
+from sqlalchemy import (
+    ARRAY,
+    JSON,
+    DateTime,
+    Float,
+    ForeignKey,
+    String,
+    Uuid,
+    create_engine,
+    select,
+    func,
+)
 from sqlalchemy.orm import DeclarativeBase, Mapped, Session, mapped_column, relationship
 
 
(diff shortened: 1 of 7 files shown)


lars-petter-hauge:
Technically we would like to get this in, but we are removing projects as well: https://github.com/equinor/acidwatch/issues/492

Closing this one, can pick it up later

PR #483, opened 2025-09-12
update Gibbs model

Sviatose:


backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -19,6 +19,7 @@ class _EquationOfState(StrEnum):
     SRK = "SRK"
     PR = "PR"
     SRKCPA = "SRKCPA"
+    IdealGas = "Ideal Gas"
```
kvashchuka, line 22: I think it will be nice to keep consistent capitalization. What about "IG" instead of "IdealGas"?

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -54,15 +56,48 @@ class GibbsMinimizationModelAdapter(BaseAdapter):
...
         "HNO3",
         "S8",
         "CH4",
-        "H2",
-        "N2",
+        "water",
+        "oxygen",
+        "sulfuric acid",
+        "nitric acid",
+        "NH4NO3",
```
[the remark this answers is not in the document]
Sviatose: Please update neqsim and solubility

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -100,38 +166,73 @@ async def run(self) -> RunResult:
...
+        system.addComponent("CO2", float(co2_content), "mole/sec")
         for component, amount in self.concentrations.items():
             neqsim_name = self.formula_to_neqsim.get(component, component)
             if amount > 0.0:
-                system.addComponent(neqsim_name, amount, "mole/sec")
-
+                system.addComponent(neqsim_name, float(amount), "mole/sec")
+        # The components below commented can be added too but only with careful evaluation of results
         system.addComponent("SO2", 0.0, "mole/sec")
         system.addComponent("SO3", 0.0, "mole/sec")
```
lars-petter-hauge, line 183: This is getting in to quite the list. It seems we are adding the same components as listed in `valid_substances`. Could I suggest something along the following lines?:

```python
        NOT_BY_DEFAULT = ["hydrogen", ...]
        for component in self.valid_substances:
            if component in NOT_BY_DEFAULT:
                continue
            system.addComponent(component, 0.0, "mole/sec")
```
This will both make it easier to see which components should not be added by default, and make it less error prone (as you don't risk having a component in either of the list that shouldn't be there (e.g. removing it from valid substances, but forgetting to remove it here.

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -100,38 +166,73 @@ async def run(self) -> RunResult:
...
+        system.addComponent("CH2O", 0.0, "mole/sec")
+        system.addComponent("C2H4O", 0.0, "mole/sec")
+        system.addComponent("C2H4", 0.0, "mole/sec")
 
         if eos in (_EquationOfState.SRK, _EquationOfState.PR):
             system.setMixingRule(2)
         elif eos == _EquationOfState.SRKCPA:
             system.setMixingRule(10)
-        else:
-            raise NotImplementedError(f"Equation of state not implemented: {eos}")
```
lars-petter-hauge, line 134: The reason such a `raise` should be included is because we want to know when things aren't working as expected. It is okay that it fails. It is in fact much much better that it fails when it should, rather than just happily ignoring things. So if we add an `_EquationOfState` in the `enum`, we want to make sure we don't forget to add how that is handled down here. A `failure` at this point should definitively be flagged, cause we have probably forgot something when writing the code...

lars-petter-hauge:
Long list of new components to play with! :)

backend/src/acidwatch_api/models/gibbs_minimization_model.py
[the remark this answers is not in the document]
Sviatose: Yes good point, but shouldnt we have a test for models? Like now its good you noticed it but if we had test it would be way safer

backend/src/acidwatch_api/models/gibbs_minimization_model.py
lars-petter-hauge: Alright, let's keep it as you made it now then :) 

(fetched the code though, and it seems there's some shifting/duplication of lines that doesn't work out. E.g. the reactor is made twice..)

backend/src/acidwatch_api/models/gibbs_minimization_model.py
lars-petter-hauge: Cheers for the change!

Looking at it now though, I have a final suggestion:

Instead of going through it twice:


```python
        for component, amount in self.concentrations.items():
            neqsim_name = self.formula_to_neqsim.get(component, component)
            if amount > 0.0:
                system.addComponent(neqsim_name, float(amount), "mole/sec")
        NOT_BY_DEFAULT = [
            "hydrogen", "N2O3", "N2O", "nitrogen", "N2H4", "COS", "methane", "ammonia"
        ]
        for component in self.valid_substances:
            if component in NOT_BY_DEFAULT:
                continue
            neqsim_name = self.formula_to_neqsim.get(component, component)
            system.addComponent(neqsim_name, 0.0, "mole/sec")
```

you could do as such:

```python
        NOT_BY_DEFAULT = [
            "hydrogen", "N2O3", "N2O", "nitrogen", "N2H4", "COS", "methane", "ammonia"
        ]
        for component in self.valid_substances:
            value = self.concentrations.get(component, 0)
            if component in NOT_BY_DEFAULT and value == 0:
                continue
            neqsim_name = self.formula_to_neqsim.get(component, component)
            system.addComponent(neqsim_name, 0.0, "mole/sec")
```

Feel free to do as you see fit

Sviatose:
> **[lars-petter-hauge](/lars-petter-hauge) ** reviewed [ Sep 16, 2025 ](#pullrequestreview-3230427017)
> [ View reviewed changes ](/equinor/acidwatch/pull/483/files/6106dbf50203d893dd3d3c39a7d212efea4969ab)

Lower damping factor will require more work from the solver but also will become more robust yes

lars-petter-hauge:
> > We have a mass balance calculation though, I don't think these component names play very nicely with it.. `n-butane`. This won't be possible to handle with current implementation. One way would be to require the components to always have their chemical name instead.. Would be awkward for the users, as it's pretty normal to rather use the more human readable ones.. @pinkwah suggestions on how to handle this in a nice way for adapters?
> > An obvious issue with this is also that e.g. `n-butane` and `i-butane` has the same chemical formula..
> 
> Yeah and also some components may have same chemical formula but different characteristics, for example the chemical formula for i-butane (isobutane) and n-butane is C₄H₁₀. They have the same molecular formula but a different structural arrangement and different properties model needs to take into account

But you could also use the non-condensed way of writing the formula:

n-Butane CH3CH2CH2CH3
iso-butane (CH3)3CH

If we only use these instead we might be alright..?

(and then it's the is it user friendly? Perhaps we could have a "chemical name" description in addition to the formula? Which is also shown..)

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -54,33 +65,55 @@ class GibbsMinimizationModelAdapter(BaseAdapter):
...
     ]
-
     # Map formulas to neqsim names
     formula_to_neqsim = {
         "H2O": "water",
-        "SO2": "SO2",
-        "SO3": "SO3",
-        "NO2": "NO2",
-        "NO": "NO",
-        "NH3": "ammonia",
```
lars-petter-hauge, line 75: was this on purpose? Seems like we no longer would rename `NH3` to `ammonia`..?

Sviatose: Yes we remove ammonia from now because this chemical is unlikely to form kinetically but we dont know maybe return in the future

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -54,33 +65,55 @@ class GibbsMinimizationModelAdapter(BaseAdapter):
...
-        "NH3",
         "H2S",
         "O2",
         "H2SO4",
         "HNO3",
         "S8",
         "CH4",
-        "H2",
-        "N2",
+        "water",
```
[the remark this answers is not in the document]
Sviatose: H2O and water no 
 H2SO4 and sulfuric acid yes, there are different parameters for acid depending literature source (accentric factor, critical properties) and we are not sure which one is correct, so I stored one parameters in H2SO4 and another in sulfuric acid so we could use both but they are not the same - should be fixed in the future.

Sviatose:
> > > We have a mass balance calculation though, I don't think these component names play very nicely with it.. `n-butane`. This won't be possible to handle with current implementation. One way would be to require the components to always have their chemical name instead.. Would be awkward for the users, as it's pretty normal to rather use the more human readable ones.. @pinkwah suggestions on how to handle this in a nice way for adapters?
> > > An obvious issue with this is also that e.g. `n-butane` and `i-butane` has the same chemical formula..
> > 
> > 
> > Yeah and also some components may have same chemical formula but different characteristics, for example the chemical formula for i-butane (isobutane) and n-butane is C₄H₁₀. They have the same molecular formula but a different structural arrangement and different properties model needs to take into account
> 
> But perhaps we could also use the non-condensed way of writing the formula:
> 
> n-Butane CH3CH2CH2CH3 iso-butane (CH3)3CH
> 
> If we only use these instead we might be alright..?
> 
> (and then it's the is it user friendly? Perhaps we could have a "chemical name" description in addition to the formula? Which is also shown..)
> 
> Edit: We even have a test to verify these names:
> 
> https://github.com/equinor/acidwatch/blob/68b32e966593eb136ea8504685ed1dc2e152d5bf/backend/tests/models/test_all_adapters.py#L35
> 
> We need to update this test, as it should fail on names such as "water".

(CH3)3CH might be possible for recognition, though not by everyone, they will wonder what is especially imaging transferring all these to structural formulas: 

<html>
<body>

MEG (Monoethylene Glycol) | C₂H₆O₂ | HO-CH₂-CH₂-OH
-- | -- | --
DEG (Diethylene Glycol) | C₄H₁₀O₃ | HO-CH₂-CH₂-O-CH₂-CH₂-OH
TEG (Triethylene Glycol) | C₆H₁₄O₄ | HO-CH₂-CH₂-O-CH₂-CH₂-O-CH₂-CH₂-OH
MEA (Monoethanolamine) | C₂H₇NO | HO-CH₂-CH₂-NH₂
MDEA (Methyldiethanolamine) | C₅H₁₅N₁O₂ | HO-CH₂-CH₂-N(CH₃)-CH₂-CH₂-OH
DEA (Diethanolamine) | C₄H₁₁N₁O₂ | HO-CH₂-CH₂-NH-CH₂-CH₂-OH
Ethane | C₂H₆ | CH₃-CH₃
Propane | C₃H₈ | CH₃-CH₂-CH₃
i-Butane (Isobutane) | C₄H₁₀ | CH₃-CH(CH₃)-CH₂-CH₃
n-Butane | C₄H₁₀ | CH₃-CH₂-CH₂-CH₃
i-Pentane (Isopentane) | C₅H₁₂ | CH₃-CH(CH₃)-CH₂-CH₂-CH₃
n-Pentane | C₅H₁₂ | CH₃-CH₂-CH₂-CH₂-CH₃
Ethylene | C₂H₄ | H₂C=CH₂
Benzene | C₆H₆ | C₆H₆ (cyclic)
Toluene | C₇H₈ | C₆H₅-CH₃
o-Xylene (1,2-Dimethylbenzene) | C₈H₁₀ | C₆H₄(CH₃)₂ (with methyl groups on adjacent carbons)


</body>
</html>

        Writing them in structural formulas is a mess, so might be better just create a proper mapping in the mass balance so that if model uses names we can just map to the number of atoms (second column) - we dont need to write structure formula then

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -19,6 +28,7 @@ class _EquationOfState(StrEnum):
     SRK = "SRK"
     PR = "PR"
     SRKCPA = "SRKCPA"
+    IdealGas = "IG"
```
kvashchuka, line 31:
```suggestion
    IG = "IG"
```

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -109,29 +143,27 @@ async def run(self) -> RunResult:
...
+            "N2O3",
+            "N2O",
+            "nitrogen",
+            "N2H4",
+            "COS",
+            "methane",
+            "ammonia",
+        ]
+        for component in self.valid_substances:
+            if component in NOT_BY_DEFAULT:
```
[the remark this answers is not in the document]
lars-petter-hauge: Yeah, I think the point in this PR is that any entry in `NOT_BY_DEFAULT` should be part of the `valid_substances`. 

Perhaps it makes more sense to do something like this:

```
NOT_BY_DEFAULT = [
            "hydrogen", "N2O3", "N2O", "nitrogen", "N2H4", "COS", "methane", "ammonia"
]
ADDITIONAL_ELEMENTS = [
        "H2O",
        "SO2",
        "SO3",
        "NO2",
]

class GibbsMinimizationModelAdapter(BaseAdapter):
    valid_substances = NOT_BY_DEFAULT + ADDITIONAL_ELEMENTS
```

Really should have a better name for these though.

lars-petter-hauge:
> > 
> > 
> > But perhaps we could also use the non-condensed way of writing the formula:
> > n-Butane CH3CH2CH2CH3 iso-butane (CH3)3CH
> > If we only use these instead we might be alright..?
> > (and then it's the is it user friendly? Perhaps we could have a "chemical name" description in addition to the formula? Which is also shown..)
> > Edit: We even have a test to verify these names:
> > https://github.com/equinor/acidwatch/blob/68b32e966593eb136ea8504685ed1dc2e152d5bf/backend/tests/models/test_all_adapters.py#L35
> > 
> > We need to update this test, as it should fail on names such as "water".
> 
> (CH3)3CH might be possible for recognition, though not by everyone, they will wonder what is especially imaging transferring all these to structural formulas: "MEG", "DEG", "TEG", "MEA", "MDEA", "DEA", "ethane", "propane", "i-butane", "n-butane", "i-pentane", "n-pentane", "ethylene", "benzene", "toluene", "o-Xylene"
> 
> Writing them in structural formulas is a mess, so might be better just create a proper mapping in the mass balance so that if model uses names we can just map to the number of atoms - we dont need to write structure formula then

"mess" might be in the eyes of the beholder though ;) I think having interchangeable ways of writing them to be more messy than writing the entire chemical formula, even if it takes more space. It also means that everyone around now has to start parsing these elements in different ways. Being strict on format makes like easier for callers.

backend/src/acidwatch_api/models/gibbs_minimization_model.py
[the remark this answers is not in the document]
lars-petter-hauge: I wonder if we could solve this by adding a paracemeter e.g. TreatH2SO4AsAcid, or something similar, and have that be the one to switch.

Also, just to be clear, there is a specific distinction within the gibbz minimization model between saying H2SO4 and 'sulfuric acid'?

backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -43,44 +53,66 @@ class GibbsMinimizationModelParameters(BaseParameters):
...
             "Peng-Robinson (PR)",
             "SRK cubic + association",
+            "Ideal Gas",
         ],
     )
 
 
 class GibbsMinimizationModelAdapter(BaseAdapter):
     valid_substances = [
-        "H2O",
```
lars-petter-hauge, line 52: Are we also removing H2O? Isn't that quite often included in experiments?

lars-petter-hauge:
Let's have a discussion about when to stop working shall we 🤣 

Screenshot 2025-09-19 at 09 19 11

lars-petter-hauge:
> I made last update hope we did not change the result. We could test the model by running the following:
> 
> P:100 T:0 H2O 30 H2S 10 SO2 10 NO2 1.5 O2 10
> 
> Model should produce NH4HSO4 1.5 ppm and S8: 0.4 ppm
> 
> Note: we do test this in neqsim as well, but there were changes to the original code as you might have observed through this pull request. I think we get the result correct, but I didnt have time to check it for the very last commit, but should be fine

We haven't discussed that much how these models should be tested. What you are describing is an end to end test (E2E). These tend to be slow and often gives false positives ( we end up testing the third party library, and if test fails, is it because the third party library changed behaviour or because of us? what do we do if it does change behaviour?...)

These kind of tests should be kept to a minimum. Some of the adapters can't be tested like this either, for instance ARCS, because it calls an API... 

What we really want to test though is the behaviour of the adapters `run` function. What we can do then is to mock the third party library, and verify that the adapter does as we expect it to.

I made a (runnable) example of how one can do this for this adapter:

```python
from acidwatch_api.models.gibbs_minimization_model import (
    GibbsMinimizationModelAdapter,
    _EquationOfState,
)
import pytest
from neqsim import jneqsim
from unittest.mock import MagicMock


class MockFluid(MagicMock):
    @staticmethod
    def getNumberOfPhases():
        return 1


class MockStream(MagicMock):
    @staticmethod
    def getFluid():
        return MockFluid()


@pytest.mark.asyncio
async def test_gibbs(monkeypatch):
    # Arrange
    mocked_system = MagicMock()
    mocked_reactor = MagicMock()

    def mock_stream(*args, **kwargs):
        return MockStream()

    def mock_system_call(*args, **kwargs):
        return mocked_system

    monkeypatch.setattr(jneqsim.thermo.system, "SystemSrkEos", mock_system_call)
    monkeypatch.setattr(jneqsim.process.equipment.stream, "Stream", mock_stream)
    monkeypatch.setattr(
        jneqsim.process.equipment.reactor, "GibbsReactor", mocked_reactor
    )

    concentrations = {"H2O": 1.1}
    parameters = {
        "temperature": 250,
        "pressure": 10,
        "equation_of_state": _EquationOfState.SRK,
    }

    # Act
    adapter = GibbsMinimizationModelAdapter(
        concentrations=concentrations, parameters=parameters, jwt_token=None
    )

    await adapter.run()

    # Assert
    assert ("water", 1.1, "mole/sec") in [
        x.args for x in mocked_system.addComponent.call_args_list
    ]
```

As you see, in the end we assert that the `addComponent` has been called with the transformed name of `H2O`

lars-petter-hauge:
> (CH3)3CH might be possible for recognition, though not by everyone, they will wonder what is especially imaging transferring all these to structural formulas: "MEG", "DEG", "TEG", "MEA", "MDEA", "DEA", "ethane", "propane", "i-butane", "n-butane", "i-pentane", "n-pentane", "ethylene", "benzene", "toluene", "o-Xylene"
> 
> Writing them in structural formulas is a mess, so might be better just create a proper mapping in the mass balance so that if model uses names we can just map to the number of atoms - we dont need to write structure formula then

luckily, my imagination is worse than  my curiosity:

```python
# reference:
# https://pubchem.ncbi.nlm.nih.gov/compound/<compound_id>

COMPOUNDS = [
    "HOCH2CH2OH",  # "MEG", 174, C2H6O2
    "(CH2CH2OH)2O"  # "DEG", 8117, C4H10O3
    "HOCH2(CH2CH2O)2CH2OH"  # "TEG", 8172 C6H14O4
    "H2NCH2CH2OH",  # "MEA", 700, C2H7NO
    "CH3N(C2H4OH)2",  # "MDEA", 7767, C5H13NO2
    "(CH2CH2OH)2NH",  # "DEA", 8113, C4H11NO2
    "CH3CH3"  # "ethane", 6324, C2H6
    "CH3CH2CH3",  # "propane",6334, C3H8
    "(CH3)2CHCH3"  # "i-butane", 6360, C4H10
    "CH3CH2CH2CH3",  # "n-butane", 7843, C4H10
    # "i-pentane",
    "CH3(CH2)3CH3",  # "n-pentane", 8003, C5H12
    "CH2=CH2",  # "ethylene", 6325, C2H4
    # "benzene", C6H6
    "C6H5CH3"  # "toluene", 1140, C7H8
    "C6H4(CH3)2",  # "o-Xylene", 7237, C8H10
]
```
Had an issue finding `benzene` and `i-pentane`.. 

(really wanted to try to use their REST api: https://pubchem.ncbi.nlm.nih.gov/docs/pug-rest, but had difficulties searching for the right compound. Got too many hits and not the ones I was interested in)

lars-petter-hauge:
merged https://github.com/equinor/acidwatch/pull/498 instead. Made sure original author was author on commits in the other pr

PR #486, opened 2025-09-18
Refactoring simulation results datatype

UsamaEquinorAFK:


frontend/src/components/LabResultsPlot.tsx
```diff
@@ -32,16 +32,14 @@ const LabResultsPlot: React.FC<LabResultsPlotProps> = ({ selectedExperiments })
                         return simulationCache[cacheKey];
                     }
                     try {
-                        const simulation = await runSimulation(
-                            filteredConcs,
-                            model.parameters && Object.keys(model.parameters).length > 0
```
lars-petter-hauge, line 37: Sure about removing the check here? Should we add parameters that the adapter doesn't need? (I wonder if perhaps the backend should fail if we try to run with parameters that the adapters hasn't asked for)

lars-petter-hauge:
Well done, separate commits and all! 

Approving this, just fix the test :) 

And I had a comment, have a look at it.

Nice pair session :)

lars-petter-hauge (approved):

PR #494, opened 2025-09-22
Add reaktoro

lars-petter-hauge:


lars-petter-hauge:
Just a draft, showing the concept.

PR #499, opened 2025-09-25
Add compounds produced in simulation results to the per-component plot

haralde:
image


lars-petter-hauge:
Would be cool to merge the labresults pr first perhaps, think that's pretty ready, and will take longer to rebase than the other way? @UsamaEquinorAFK

UsamaEquinorAFK:
@lars-petter-hauge i have merged now :)

PR #501, opened 2025-09-25
Add tooltip to settings input

haralde:
Adds tootip to model parameter, text is taken from model parameter description. Might need to update some of the text

image


lars-petter-hauge:
@UsamaEquinorAFK has a similar PR, no?

UsamaEquinorAFK (approved):
lgtm

PR #511, opened 2025-10-02
Enable downloading of graph data from LabResult page

UsamaEquinorAFK:


frontend/src/components/DownloadButton.tsx
```diff
@@ -0,0 +1,35 @@
+﻿import React from "react";
+import { Button, Icon } from "@equinor/eds-core-react";
+import { save } from "@equinor/eds-icons";
+import { ChartDataSet } from "../dto/ChartData";
+
+Icon.add({ save });
```
pinkwah, line 6: This seems like a smell (modifying global state just by importing this module)

Maybe you should change the usage of `<Icon>` to have `data`? Ie, `<Icon data={save} />`.

lars-petter-hauge:
solves #511

UsamaEquinorAFK:
> when the data is downloaded we also include all components that don't have a value. We should remove them and only include the ones with a number.
> 
> technically we would like to only show the ones where at least one row has a value greater than 0, with the special case that we want to show output columns even with all 0, with the input column does have value greater than 0 (so if one starts with 10 ppm H2S, the 0 ppm H2S in the output has a meaning)

Fixed :))

PR #524, opened 2025-10-17
Convert to json records format in backend and use generic table compnent.   

haralde:


backend/src/acidwatch_api/models/arcs.py
```diff
@@ -84,9 +84,28 @@ async def run(self) -> RunResult:
         )
 
         result = response.json()
+        paths = result["analysis"]["common_paths"]
+        stats = result["analysis"]["stats"]
+        common_paths = [
+            {
+                "Path": v.replace("<sub>", "").replace("</sub>", ""),
```
lars-petter-hauge: So this would turn `H<sub>2</sub>O` into `H2O`. We haven't really landed on how to treat these chemical formulas. I think other adapters would respond with `H2O`, and it would be nice to be consistent on this. Also, we treat all input components as `H2O`.

This kinda breaks a bit the backend api, always nice to keep in mind. But for the better.

backend/src/acidwatch_api/models/arcs.py
```diff
@@ -84,9 +84,28 @@ async def run(self) -> RunResult:
...
+            }
+            for k, v in paths["paths"].items()
+        ]
+        all_stats = [
+            {
+                "Path": v,
+                "k": stats["k"][k],
+                "Frequency": stats["frequency"][k],
+            }
+            for k, v in stats["index"].items()
```
lars-petter-hauge: yeah, this just makes more sense, 

so we are going from:
 
```python
{'index': {'0': '4 HNO3 = 2 H2O + 4 NO2 + O2', '1': '2 H2SO4 = 2 H2O + O2 + 2 SO2', }, 'k': {'0': ' 1.35e-33', '1': ' 7.24e-30', '2': ' 5.93e-11', }, 'frequency': {'0': 739, '1': 557,}}
```

to:
```python
[
{'Path': '4 HNO3 = 2 H2O + 4 NO2 + O2', 'k':  '1.35e-33', 'frequency': 739},
{'Path': '2 H2SO4 = 2 H2O + O2 + 2 SO2', 'k': '7.24e-30', 'frequency': 557}
]
```

backend/src/acidwatch_api/models/arcs.py
haralde: Yes, other adapters and even other data from arcs returns H2O, and then this will again be converted to correct subscripts in frontend. I would like to remove the formatting in arcs on that part, so that we are consistent. This is kind of temprary

PR #534, opened 2025-10-21
Implement async simulation and results

pinkwah:


backend/tests/test_models_endpoints.py
```diff
@@ -109,8 +119,16 @@ def test_dummy_model_only_valid_substances_are_present(
         }
     else:
         response.raise_for_status()
+        simulation_id = response.json()
+
+        response = client.get(f"/simulations/{simulation_id}/result")
```
lars-petter-hauge: do we expect that this could turn out to be a flaky test? feels like it

have a function that will retry on "pending" response?

frontend/src/contexts/SimulationContext.tsx
```diff
@@ -14,27 +15,23 @@ const SimulationResultsContext = createContext<SimulationResultsContextType>(nul
...
+        queryFn: async () => startSimulation(modelInput!),
+        enabled: modelInput !== undefined,
+    });
+
+    const { data: simulationResults, isLoading } = useQuery({
+        queryKey: ["simulationResult", simulationId],
+        queryFn: async () => getResultForSimulation(simulationId!),
+        retry: (_failureCount, error) => error instanceof ResultIsPending,
+        enabled: simulationId !== undefined,
+    });
```
lars-petter-hauge: This looks quite nice really

frontend/src/contexts/SimulationContext.tsx
```diff
@@ -14,27 +15,23 @@ const SimulationResultsContext = createContext<SimulationResultsContextType>(nul
 
 export const SimulationResultsProvider: React.FC<{ children: ReactNode }> = ({ children }) => {
     const [modelInput, setModelInput] = useState<ModelInput | undefined>(undefined);
-    const [simulationResults, setSimulationResults] = useState<SimulationResults | undefined>(undefined);
-    const [loading, setLoading] = useState(false);
```
lars-petter-hauge: remove state, excellent! and just use `useQuery` state instead

frontend/src/dto/SimulationResults.tsx
```diff
@@ -27,6 +27,7 @@ interface TablePanel {
 
 export type Panel = TextPanel | JsonPanel | ReactionPathsPanel | TablePanel;
 export interface SimulationResults {
+    status: "error" | "pending";
```
lars-petter-hauge: should it not be `done` instead of `error`?

pinkwah:
`UseSimulation` will be called _many_ times, but the fetch should happen once per simulation.

What you're seeing is the result of some queries completing and thus partial results becoming available. The log statement you added isn't for fetching, but for creating the queries. The results of the queries are all cached, but to discover this you need call the closure where you added the log a bunch of times.

It may seem weird to re-run these functions over and over, but this is how React works.

PR #577, opened 2025-11-14
Remove unused projects and depending functionality

UsamaEquinorAFK:


lars-petter-hauge:
cool, can remove switchPublicity and saveProject as well, right?

Issue #596, opened 2025-12-01
Add concentration inputs when going to a simulation page


lars-petter-hauge:
currently, if a user runs a simulation and then visits the page including the simulation id, the input will be gone.

The simulationresults does however contain the values. Set the concentrations according to the Simulationresults for the given model and set it as current model.

PR #597, opened 2025-12-01
Populate modelInput with sim result input

lars-petter-hauge:
solves: #596

~fixes concentrations, but not parameters, those are slightly more tricky~

Will not interfere with any of the input concentrations from other models

### DIFF
--- a/frontend/src/components/ModelInputs.tsx
+++ b/frontend/src/components/ModelInputs.tsx
@@ -105,14 +105,23 @@ const ModelInputs: React.FC<{
     model: ModelConfig;
     visible: boolean;
     onSubmit: (concentrations: Record<string, number>, parameters: Record<string, number>) => void;
-}> = ({ model, visible: not_hidden, onSubmit }) => {
-    const filteredDefaults = Object.fromEntries(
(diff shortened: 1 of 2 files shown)


lars-petter-hauge:
tried handling this with using `key` attribute to a larger degree:

```typescript
        inputsStep = models.map((model) => {
            const isCurrentModel = model.modelId === currentModel?.modelId;

            const defaultConcentrations = isCurrentModel ? simulationResults?.modelInput?.concentrations : undefined;
            const defaultParameters = isCurrentModel ? simulationResults?.modelInput?.parameters : undefined;

            return (
                <ModelInputs
                    key={`${model.modelId}-${JSON.stringify(defaultConcentrations)}-${JSON.stringify(defaultParameters)}`}
                    model={model}
                    visible={isCurrentModel}
                    onSubmit={(concentrations, parameters) =>
                        setModelInput({ modelId: model.modelId, concentrations, parameters })
                    }
                    defaultConcentrations={defaultConcentrations}
                    defaultParameters={defaultParameters}
                />
            );
        });
```

but getting issue where when a user press `run`, the numbers in the input form goes back to default values, before resetting to simulation results when the result is retrieved..

(also all input models all of a sudden has the same concentrations (eg invalid components are listed..))

PR #605, opened 2025-12-09
Reload page when using navigation buttons to ensure correct inputs

### LINKED ISSUE #599: Bug whenever using navigation in simulation tab
The modelinput for a specific model does not update when using navigation (backwards and forwards)

see: https://github.com/equinor/acidwatch/pull/597#issuecomment-3597253331

haralde:
Fixes #599 by forcing a refresh of entire page when using navigation buttons.
Not sure if this is the correct reacty way of doing it but it works


pinkwah:
#615 makes this not necessary. Closing (but reopen if I goofed)

PR #623, opened 2025-12-18
Subsequent Model runs 

UsamaEquinorAFK:


frontend/src/functions/Filtering.ts
```diff
@@ -9,3 +9,44 @@ export const filterValidModels = (experiment: ExperimentResult, models: ModelCon
...
+        )
+        .reduce(
+            (acc, [substance, concentration]) => ({
+                ...acc,
+                [substance]: Number(concentration),
+            }),
+            {} as Record<string, number>
+        );
+}
+/* Specific for Solubility CCS model - added after consultation with domain experts*/
```
haralde: Move model specific stuff to the backend?

PR #686, opened 2026-04-30
Use Equinor colors for plots

haralde:


frontend/src/functions/Colors.ts
```diff
@@ -1,5 +1,24 @@
...
+] as const;
+
+export const getDistributedColor = (current: number, total: number): string => {
+    if (total <= 0) return EQUINOR_CHART_PALETTE[0];
+
+    if (total <= EQUINOR_CHART_PALETTE.length) {
+        return EQUINOR_CHART_PALETTE[current % EQUINOR_CHART_PALETTE.length];
+    }
+
+    const scale = chroma.scale(EQUINOR_CHART_PALETTE as unknown as string[]).mode("lab");
```
lars-petter-hauge: This double casting is a bit off, don't you think? I understand it's because of the readonly list of strings. A possible way of doing it could be to unpack in to a new list and provide that instead:

`chroma.scale([...EQUINOR_CHART_PALETTE]).mode("lab");`

PR #687, opened 2026-04-30
Separate temperature and pressure from model parameters (#636)

### LINKED ISSUE #636: Separate conditions (temperature and pressure)
It doesn't make sense to have separate conditions for reactive and solubility models. The defaults are different, set by the adapter. This means that if the user forgets to change to the same conditions, it will be incorrect.

Conditions are describing the system (same with concentrations), while other parameters are tuning the models.

Have conditions be on the same level as concentrations.

Note that models have different minimum and maximum on the conditions. Consider that the range must then be the union of conditions for all selected models.

haralde:
Closes #636.

Lifts `temperature` and `pressure` out of each adapter's parameters
schema and into top-level simulation conditions, so they're entered
once and shared across primary and secondary models.

### Backend
- `BaseAdapter` accepts `temperature`/`pressure` kwargs and validates
  them against optional `temperature_range`/`pressure_range` class
  attributes; validation errors surface under a new `conditions` key.
- T/P removed from `*Parameters` classes in `arcs`, `arcs_exp`,
  `gibbs_minimization_model`, `phpitz`, `solubilityccs`. Each adapter
  now declares its supported ranges. ToCoMo unchanged (no T/P).
- `Simulation` and `ModelInfo` Pydantic models gain `temperature`,
  `pressure`, `temperature_range`, `pressure_range`.
- `simulations` table gains nullable `temperature`/`pressure` columns.
- Alembic migration `7c2e9a4b1f3d_separate_conditions` lifts existing
  values out of `model_inputs.parameters` JSON into the new columns.

### Frontend
- New `useConditionsStore` holds the shared T/P values.
- `ModelInputs` renders a Conditions block below Concentrations, using
  the **intersection** of the selected models' supported ranges. The
  block is hidden when no selected model needs T/P (e.g. ToCoMo only),
  in which case `null` is sent to the API.
- `Models` page restores T/P when reloading a historical simulation.
- DTOs (`ModelInput`, `FormConfig`) extended with the new fields.

### DIFF
--- a/backend/src/acidwatch_api/models/base.py
+++ b/backend/src/acidwatch_api/models/base.py
@@ -195,8 +195,39 @@ class BaseAdapter:
         *,
         concentrations: dict[str, int | float] | None = None,
         parameters: dict[str, str | bool | int | float] | None,
+        temperature: float | None = None,
+        pressure: float | None = None,
         jwt_token: str | None,
     ) -> None:
+        condition_errors: dict[str, list[str]] = {}
+        temperature_range = type(self).temperature_range
+        if temperature_range is not None:
+            if temperature is None:
+                condition_errors["temperature"] = ["Field required"]
+            elif not (temperature_range[0] <= temperature <= temperature_range[1]):
+                condition_errors["temperature"] = [
+                    f"Input must be between {temperature_range[0]} and {temperature_range[1]}"
+                ]
+        pressure_range = type(self).pressure_range
+        if pressure_range is not None:
+            if pressure is None:
+                condition_errors["pressure"] = ["Field required"]
+            elif not (pressure_range[0] <= pressure <= pressure_range[1]):
+                condition_errors["pressure"] = [
+                    f"Input must be between {pressure_range[0]} and {pressure_range[1]}"
+                ]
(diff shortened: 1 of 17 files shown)


backend/src/acidwatch_api/models/base.py
```diff
@@ -195,8 +195,39 @@ def __init__(
...
     ) -> None:
+        condition_errors: dict[str, list[str]] = {}
+        temperature_range = type(self).temperature_range
+        if temperature_range is not None:
+            if temperature is None:
+                condition_errors["temperature"] = ["Field required"]
+            elif not (temperature_range[0] <= temperature <= temperature_range[1]):
+                condition_errors["temperature"] = [
+                    f"Input must be between {temperature_range[0]} and {temperature_range[1]}"
+                ]
```
lars-petter-hauge, line 210: My immediate thought here is that we have the Parameter made to handle Min and Max etc, while this implementation creates its own set of validation by considering it as a tuple range. Neither is really wrong, but then we end up with two different implementations. Frontend is also set up to handle the `Parameter` type. Could we try to re-use the `Parameter`?

Issue #635, opened 2026-01-07
Remove the option of changing between units (kelvin and celcius)


lars-petter-hauge:
Keep only celcius. It adds complexity to handle this. If we change our minds we can add it again later.

PR #688, opened 2026-05-06
Separate temperature and pressure from model parameters (#636)

haralde:
Lifts temperature and pressure out of each model adapter's parameters schema and into top-level simulation conditions, so they're entered once and shared across primary and secondary models.

No need to specify default temp and pressure for each model

### DIFF
--- a/backend/src/acidwatch_api/models/datamodel.py
+++ b/backend/src/acidwatch_api/models/datamodel.py
@@ -19,8 +19,14 @@ class ModelInput(_BaseModel):
     parameters: dict[str, bool | float | int | str]
 
 
+class Conditions(_BaseModel):
+    temperature: float = 300
+    pressure: float = 10
+
+
 class Simulation(_BaseModel):
     concentrations: dict[str, int | float]
+    conditions: Conditions = Field(default_factory=Conditions)
     models: list[ModelInput] = Field(min_length=1)
 
 
--- a/backend/src/acidwatch_api/models/gibbs_minimization_model.py
+++ b/backend/src/acidwatch_api/models/gibbs_minimization_model.py
@@ -145,7 +145,7 @@ class GibbsMinimizationModelAdapter(BaseAdapter):
     async def run(self) -> RunResult:
         eos = self.parameters.equation_of_state
         temperature = self.conditions.temperature
-        pressure= self.conditions.pressure
+        pressure = self.conditions.pressure
 
         if eos == _EquationOfState.SRK:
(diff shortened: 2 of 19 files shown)


backend/src/acidwatch_api/models/gibbs_minimization_model.py
```diff
@@ -191,8 +176,8 @@ async def run(self) -> RunResult:
 
         # # Create an inlet stream
         inlet_stream = jneqsim.process.equipment.stream.Stream("Inlet Stream", system)
-        inlet_stream.setPressure(self.parameters.pressure, "bara")
-        inlet_stream.setTemperature(self.parameters.temperature, "K")
+        inlet_stream.setPressure(self.conditions.pressure, "bara")
```
kvashchuka: Can't we use `temp` and `pres` variables here?

kvashchuka: Which btw is a misleading choice of name, `temp` is usually something temporary. I vote for spelling out `pressure` and `temperature`

backend/src/acidwatch_api/models/datamodel.py
```diff
@@ -19,8 +19,14 @@ class ModelInput(_BaseModel):
     parameters: dict[str, bool | float | int | str]
 
 
+class Conditions(_BaseModel):
```
kvashchuka, line 22: I guess we will now assume units of K for temperature and bara for pressure? I understand that each model should not care about it, but maybe it makes sense to set units and min max in the backend? If I understood correctly now it is set when rendering in frontend

kvashchuka (commented):
left some comments

backend/src/acidwatch_api/models/gibbs_minimization_model.py
kvashchuka: I see that it was like this from before, but I would still change it while we are at it

kvashchuka:
Btw I tried locally just now and the simulations do not run

pinkwah:
Can you add tests for the migration? Ie, have results with t&p in parameters, then perform migration. Also have it for the downgrade.

haralde:
>  Btw I tried locally just now and the simulations do not run

Running Arcs from main locally does not work, need to run from an old commit. This is because of the setup we have with arcs vs arcs_experimental. We should re-think this, it quickly becomes a mess when trying to keep track of this

backend/src/acidwatch_api/models/datamodel.py
haralde: This will be handled in a separate PR (#635)

PR #695, opened 2026-05-11
Restrict HTTP methods and headers with CORS

fmell:
Restrict allowed CORS methods to: `GET`, `POST`, `PUT`, `DELETE`, `OPTIONS`
Restrict allowed CORS headers to: `Authorization`, `Content-Type`
Add fail-fast validation that raises on origin `"*"` when `allow_credentials` is true, as this will fail silently in the browser if misconfigured


haralde (approved):

lars-petter-hauge:
don't see a reason for not adding this, gogogo :)

Issue #701, opened 2026-05-15
Colors in lab results plot


kvashchuka:
I suggest we use Equinor colors from the color palette for graphs and infographics [link](https://communicationtoolbox.equinor.com/point/en/equinor/component/default/100062).

I think we could use North sea (dark blue) and it's shades for first experiment and corresponding simulation model results, Norwegian wood (green) and it's shades for the next, Sand and summer (red) for the third. These three colors have the most shades available, which should be enough to accommodate many simulation models results. We need to decide what to do with the rest - https://github.com/equinor/acidwatch/issues/700

Idea: explore potential to add stripes or shapes to the bars to ease distinguishing between experiments

Something to be aware of: we should keep the three colors for the `Output concentrations` plots in `Models` to  the North sea, Norwegian Woods and Autumn Storm (blue, green, gray). Right now they are the top three color from EQUINOR_CHART_PALETTE. The colors for the lab results bar plot are also taking in order from the EQUINOR_CHART_PALETTE

Issue #700, opened 2026-05-15
Discuss the color palette for the lab results graph


kvashchuka:
I want to migrate to Equinor colors in all plots. As of now, lab results page all for selection of as many lab results as you want , and it will put all of them in one graph.

It gets busy quite fast, so I do not think people choose more than 3, but this believe needs to be verified with users.

I suggest we have discussion with extended group of SMEs to decide either to limit the number of experiments we allow to pick, or figure out the colors

PR #702, opened 2026-05-18
Enhance visibility in lab results plot

haralde:
#701 

Using only 3 main EQ colors for experiments, and shadings for model runs.
Added stripes to bars and sorted and moved legends to the right.

Stripes might not be needed when bars and legends are sorted?

image


frontend/tests/components/LabResultPlot.test.tsx
```diff
@@ -123,10 +123,7 @@ describe("LabResultsPlot Component", () => {
             { x: "H2O", y: 0.4 },
         ]);
 
-        expect(graphData[1].data).toEqual([
-            { x: "CO2", y: 0.3 },
-            { x: "H2O", y: 0.5 },
-        ]);
```
lars-petter-hauge, line 129: should this test change due to the changes in coloring?

lars-petter-hauge:
silly question, but pattern changes per experiment. We have `BarPattern = "solid" | "diagonal" | "crosshatch" | "dots";
`. What happens when the fifth experiment is selected?

haralde:
> silly question, but pattern changes per experiment. We have `BarPattern = "solid" | "diagonal" | "crosshatch" | "dots"; `. What happens when the fifth experiment is selected?

Patterns will be recycled. But might limit to 3 experiments anyway (#700)

PR #704, opened 2026-05-20
Remove pickletype

fmell:


haralde (approved):

backend/src/acidwatch_api/routes/models.py
```diff
@@ -126,18 +128,23 @@ async def _run_adapter(
...
             concentrations=concs,
             panels=[p.model_dump(mode="json", by_alias=True) for p in panels],
-            python_exception=None,
             error=None,
         )
 
         return concs
     except BaseException as exc:
+        # Full traceback goes to logs (App Insights); only a short message
+        # is persisted for surfacing to the API caller.
```
lars-petter-hauge, line 137: question, the ModelResult part is the information that will be returned to the caller, this log statement though, will only be found in the terminal starting the backend, correct?

frontend/src/api/api.ts
```diff
@@ -63,11 +63,13 @@ async function apiRequest<Model extends z.ZodTypeAny>(
...
-    const headers = {
+    const headers: Record<string, string> = {
         "Content-Type": "application/json",
-        Authorization: `Bearer ${token}`,
-        ...(init.headers ?? {}),
+        ...Object.fromEntries(new Headers(init.headers as HeadersInit)),
     };
+    if (token) {
+        headers["Authorization"] = `Bearer ${token}`;
+    }
```
lars-petter-hauge, line 72: Cheers :)

backend/src/acidwatch_api/routes/models.py
haralde: Yes, and in app-insights hopefully. We could not verify this locally, because I could not connect to appinsights from my machine. Lets see what happens in dev

PR #709, opened 2026-05-22
Disable editable uv sync in Dockerfile

fmell:


backend/entrypoint
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 set -ex
 
-uv run alembic upgrade head
-exec uv run uvicorn acidwatch_api.app:app --host 0.0.0.0 --port 8000
+uv run --no-sync alembic upgrade head
```
lars-petter-hauge, line 4: is this to speed up the process? Since the dockerfile already has `sync --frozen` we know the environment is up to date and we don't need to check again?

edit: Just read that by default for `uv run` commands `uv` will in fact update the lockfile. I understand.

Did not know `uv` behaved like that for `uv run`, learn something new!

PR #714, opened 2026-06-08
Remove N2O from Gibbs Model

Sviatose:


kvashchuka:
Hi @Sviatose! Could you write down some reasoning behind this decision?

kvashchuka:
I would also like to the see the commit message same as the title of this PR:) It is very helpful for future us if the git history is clear

Sviatose:
When I test H2O 100 ppm H2S 100 ppm SO2 100 ppm NO 100 ppm the model produces a lot of N2O which I believe is not correct and instead ammonia should form. I think we should do experiments and confirm which products are formed to be sure, but I dont think we can form N2O in so high amounts

kvashchuka (approved):

PR #741, opened 2026-06-16
Enforce single result per modelinput

lars-petter-hauge:
In Acidwatch we allow for a simulation consisting of multiple model inputs (chaining models), and for each modelinput there corresponds a modelresult. This was added as part of the orm, but we should enforce it also on the db level.


lars-petter-hauge:
I don't see a reason for why it wouldn't be true. If there in fact does exist multiple results with the same modelinput, the migration will fail and we will find out that I was wrong -.-

PR #756, opened 2026-06-19
Update results tables to handle phases

lars-petter-hauge:
Currently the result tables are off, as they only consider a single phase:

The mass transfer between them is lost 

Screenshot 2026-06-19 at 10 23 37
Screenshot 2026-06-19 at 10 23 43


lars-petter-hauge:
new suggestion:

Screenshot 2026-06-19 at 10 39 41

PR #776, opened 2026-06-23
Update grid sim result

lars-petter-hauge:


lars-petter-hauge:
I see some places where we can improve (at least some complexity reduction), but otherwise this PR touches little but the grid sections (the refactoring in phaseresult and results are minimal). I will merge this PR and try to improve a bit afterwards.

Issue #804, opened 2026-08-01
Simplified architecture with message broker and containerised models


lars-petter-hauge:
Solving https://github.com/equinor/acidwatch/issues/809

I think I have a suggestion for a simplified design, that utilises a messagebroker to a larger degree for keeping state (that is, we can derive state), and make it drive the orchestration. This will, I think, lead to less test cases and easier test setup.

Main part is, the first api call from the user will submit the first message in the chain, and then return. Fire and forget. Then have a listener on the backend side which will update database with results and submit another job if necessary.

Pretty much everything in this setup is fire and forget.

What follows is a range of flow diagrams for various scenarios, hopefully covered most.

In these flow diagrams the result listener and api seems to be two different interfaces. I am uncertain where the distinction between them will be. I was kinda hoping for just two asyncio loops in the backend, on for the api and one for the result listener (and the heartbeat, shown later).

### **Scenario 1: Happy Path (Normal Execution)**

```mermaid
sequenceDiagram
    participant Client
    box Backend
    participant API as Web Api
    participant Listener as Result Listener
    end
    participant DB as Database
    participant Broker as Message Broker
    participant Worker_A as Worker (Model A)
    participant Worker_B as Worker (Model B)

    Client->>API: POST /simulations<br/>(models=[A, B])
    activate API
    API->>DB: Create Simulation<br/>+ ModelInputs (A → B chain)
    activate DB
    DB-->>API: simulation_id
    deactivate DB
    API->>Broker: Publish Job(A)<br/>to acidwatch.model_a
    activate Broker
    Broker-->>API: Queued
    deactivate Broker
    API-->>Client: Return simulation_id<br/>(FIRE & FORGET)
    deactivate API

    Broker->>Worker_A: Deliver Job(A)
    activate Worker_A
    Worker_A->>Worker_A: Run Model A<br/>concentrations → results
    Worker_A->>Broker: Publish result to<br/>acidwatch.results
    deactivate Worker_A

    Broker->>Listener: Deliver result
    activate Listener
    Listener->>DB: Insert ModelResult(A)<br/>with phases/error
    activate DB
    DB-->>Listener: ✓ Saved
    deactivate DB
    Listener->>DB: Query next model<br/>after A → find B
    activate DB
    DB-->>Listener: ModelInput(B)
    deactivate DB
    Listener->>Broker: Publish Job(B)<br/>to acidwatch.model_b
    Listener->>Broker: Ack result message
    deactivate Listener

[...]

PR #807, opened 2026-08-04
Messagebroker feature

lars-petter-hauge:
Implementing https://github.com/equinor/acidwatch/issues/804

Looks like a lot of changes in this one, but a lot of them are from uv files as each model has their entirely own setup. Also considering that it is a major change in architecture, it isn't as bad as it looks.

```
git diff --stat origin/main -- . ':(exclude)*uv.lock' ':(exclude)*Docker*' ':(exclude)*docker*' ':(exclude)*radix*'
...
 118 files changed, 4286 insertions(+), 878 deletions(-)
```

Also excluding test files:

```
git diff --stat origin/main -- . ':(exclude)*uv.lock' ':(exclude)*Docker*' ':(exclude)*docker*' ':(exclude)*radix*' ':(exclude)*test*'
...
 82 files changed, 2065 insertions(+), 820 deletions(-)
```

Also, I do not intend to sneak this one in https://github.com/equinor/acidwatch/pull/807/changes/34951dd11a5aa7c039cbf3153f11074353fdb1f5

but we haven't required authentication for any models at all. I think we should consider this later if necessary.. 

A lot of these commits are preparing for change. It is really this commit that changes how the backend works, and which will then rely on a messagebroker in order to function: **https://github.com/equinor/acidwatch/pull/807/changes/7960b7b53948dc5117a1a4a18692b6533fae04e7**

Until this part I am actually adding packages to acidwatch backend (that is, those worker projects), and then in this commit removing all those dependencies again. I think this make the git history the most clean and easily readable (fingers crossed..)


backend/packages/acidwatch-messaging/src/acidwatch_messaging/transport.py
```diff
@@ -0,0 +1,203 @@
...
+        exc_type: type[BaseException] | None,
+        exc: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        if self._connection is not None:
+            await self._connection.close()
+
+    async def _declare_queues(self, channel: aio_pika.abc.AbstractChannel) -> None:
+        if DEAD_LETTER_QUEUE in self._queue_names:
+            await channel.declare_queue(DEAD_LETTER_QUEUE, durable=True)
```
lars-petter-hauge, line 77: Is this because empty args is not allowed? Otherwise I don't see a reason to special case this queue

backend/packages/acidwatch-messaging/src/acidwatch_messaging/transport.py
```diff
@@ -0,0 +1,203 @@
...
+            aio_pika.Message(
+                body=_payload_json(payload).encode(),
+                delivery_mode=aio_pika.DeliveryMode.PERSISTENT,
+                content_type="application/json",
+            ),
+            routing_key=queue_name,
+            mandatory=True,
+        )
+
+    async def subscribe(self, queue_name: str) -> AsyncIterator[Message]:
```
lars-petter-hauge: Theres something in me saying that for msg in instance.subscribe() isn't obvious way to use a subscribe.. is that the most common way? To get an iterator on a function named "subscribe"?

backend/packages/acidwatch-messaging/src/acidwatch_messaging/transport.py
```diff
@@ -0,0 +1,203 @@
...
+    async def reject(self) -> None: ...
+
+
+class Transport(Protocol):
+    async def publish(self, queue_name: str, payload: Payload) -> None: ...
+
+    def subscribe(self, queue_name: str) -> AsyncIterator[Message]: ...
+
+
+class RabbitMQMessage:
```
lars-petter-hauge, line 41: Should these classes not show that they implement the protocols declared above?

backend/packages/acidwatch-models/src/acidwatch_models/base.py
```diff
@@ -201,6 +199,7 @@ def __init__(
         parameters: dict[str, str | bool | int | float] | None,
         conditions: Conditions | None = None,
         jwt_token: str | None,
+        access_token: str | None = None,
```
lars-petter-hauge: Consider dropping auth altogether here, we haven'r been using auth at all. Pr upfront would be best

backend/src/acidwatch_api/models/__init__.py
```diff
@@ -1,21 +1,17 @@
-from .base import BaseAdapter, InputError, get_parameters_schema
```
lars-petter-hauge, line 1: Lets not change sorting in import like this in the same commit.

backend/packages/acidwatch-models/src/acidwatch_models/definitions/arcs_exp.py
```diff
@@ -1,11 +1,10 @@
-from acidwatch_api.models.base import (
-    BaseAdapter,
-    RunResult,
-)
-from acidwatch_api.models.datamodel import Phase
-from acidwatch_api.settings import SETTINGS
+from acidwatch_models.base import BaseAdapter
 
-DESCRIPTION: str = """\
```
lars-petter-hauge, line 8: No need to change this part in the same commit

backend/packages/acidwatch-models/src/acidwatch_models/definitions/example.py
```diff
@@ -0,0 +1,21 @@
+from acidwatch_models import BaseAdapter, BaseParameters, Parameter
+
+
+class ExampleParameters(BaseParameters):
```
lars-petter-hauge, line 4: I would have expected this to be a file move instead of creating a new file. I think a rebase where we first move the adapter files into the definitioms folder, and then remove the run functionality in a separate commit, would make this readable

backend/packages/acidwatch-models/src/acidwatch_models/definitions/tocomo.py
```diff
@@ -1,12 +1,10 @@
-from __future__ import annotations
-from acidwatch_api.models.base import BaseAdapter, RunResult
-from acidwatch_api.models.datamodel import Phase, TableResult
-from fastapi import APIRouter
-from acidwatch_api.settings import SETTINGS
+from acidwatch_models.base import BaseAdapter
 
-router = APIRouter()
 
-DESCRIPTION: str = """\
```
lars-petter-hauge, line 9: Another uneccessary change..

backend/packages/acidwatch-worker-runtime/src/acidwatch_worker_runtime/runtime.py
```diff
@@ -0,0 +1,156 @@
...
+        *,
+        heartbeat_interval: float = 30,
+    ) -> None:
+        self._adapter_type = adapter_type
+        self._transport = transport
+        self._heartbeat_interval = heartbeat_interval
+        self._instance_id = str(uuid4())
+        self._job_id: UUID | None = None
+
+    async def run_job(self, job: AdapterJob) -> AdapterResult:
```
lars-petter-hauge: Shouldn't one of run and run_job be private? Otherwise a user of this class could be confused. (I mean, I am)

backend/packages/acidwatch-worker-runtime/src/acidwatch_worker_runtime/runtime.py
```diff
@@ -0,0 +1,156 @@
...
+        RESULTS_QUEUE,
+        HEARTBEATS_QUEUE,
+        DEAD_LETTER_QUEUE,
+    ]
+    transport = create_transport(broker_url, queues, backend)
+    async with transport:
+        await AdapterWorker(adapter_type, transport).run()
+
+
+def run_worker_from_environment(adapter_type: type[BaseAdapter]) -> None:
```
lars-petter-hauge, line 171: Is this wrapper neccessary?

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,128 @@
...
+    RESULTS_QUEUE,
+    Transport,
+    job_queue_name,
+)
+from acidwatch_models.datamodel import Conditions, Phase
+
+logger = logging.getLogger(__name__)
+
+
+async def supervise_listener(
```
lars-petter-hauge: What's this?

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,128 @@
...
+    async for message in transport.subscribe(HEARTBEATS_QUEUE):
+        try:
+            heartbeat = Heartbeat.model_validate(message.body)
+            registry.update(
+                heartbeat.model_id,
+                instance_id=heartbeat.instance_id,
+                timestamp=heartbeat.timestamp,
+                job_id=heartbeat.job_id,
+            )
+        except ValidationError:
```
lars-petter-hauge, line 99: Is the validationerror from the heartbeat.model_validate only, or also on register.update? Should we move the except one up?

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,128 @@
...
+            continue
+
+        next_job: tuple[str, AdapterJob] | None = None
+        with sessionmaker() as session:
+            model_input = session.get(
+                db.ModelInput,
+                incoming_result.model_input_id,
+            )
+            if model_input is None:
+                await message.reject()
```
lars-petter-hauge: How could we get into this scenario? Invalid model_input.id I suppose?

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,128 @@
...
+                continue
+
+            existing_result = session.scalar(
+                select(db.ModelResult).where(
+                    db.ModelResult.model_input_id == incoming_result.model_input_id
+                )
+            )
+            if existing_result is None:
+                existing_result = db.ModelResult(
```
lars-petter-hauge: It is not existing result here, its about to be committed.. rename

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,128 @@
...
+            if model_input is None:
+                await message.reject()
+                continue
+
+            existing_result = session.scalar(
+                select(db.ModelResult).where(
+                    db.ModelResult.model_input_id == incoming_result.model_input_id
+                )
+            )
+            if existing_result is None:
```
lars-petter-hauge: Could we do if existing_result: continue instead?

Might need to publish next...

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,128 @@
...
+        except ValidationError:
+            await message.reject()
+            continue
+        await message.ack()
+
+
+async def result_listener(transport: Transport, sessionmaker: db.SessionMaker) -> None:
+    async for message in transport.subscribe(RESULTS_QUEUE):
+        try:
+            incoming_result = AdapterResult.model_validate(message.body)
```
lars-petter-hauge: A bit difficult knowing what is referring to broker objects and what is referring to db objects. Maybe we could make names that fit with their origin

backend/src/acidwatch_api/database.py
```diff
@@ -116,10 +119,68 @@ async def lifespan(_: FastAPI) -> AsyncIterator[AppState]:
     state: AppState = {
         "engine": engine,
         "session": session,
+        "transport": None,
+        "heartbeat_registry": None,
     }
 
     try:
-        yield state
+        from acidwatch_api.broker.heartbeat import HeartbeatRegistry
```
lars-petter-hauge: Move to top

backend/src/acidwatch_api/database.py
```diff
@@ -116,10 +119,68 @@ async def lifespan(_: FastAPI) -> AsyncIterator[AppState]:
...
+                supervise_listener,
+            )
+            from acidwatch_messaging import (
+                DEAD_LETTER_QUEUE,
+                HEARTBEATS_QUEUE,
+                RESULTS_QUEUE,
+                create_transport,
+                job_queue_name,
+            )
+            from acidwatch_models import get_adapters
```
lars-petter-hauge: Why are all of these so far down?

Are we considering that there could be different communication channels? I think we should accept the broker pattern completely

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,128 @@
...
+                timestamp=heartbeat.timestamp,
+                job_id=heartbeat.job_id,
+            )
+        except ValidationError:
+            await message.reject()
+            continue
+        await message.ack()
+
+
+async def result_listener(transport: Transport, sessionmaker: db.SessionMaker) -> None:
```
lars-petter-hauge, line 111: This function does not read very well, would be nice to refactor it a bit

backend/src/acidwatch_api/database.py
```diff
@@ -116,10 +119,68 @@ async def lifespan(_: FastAPI) -> AsyncIterator[AppState]:
...
         "session": session,
+        "transport": None,
+        "heartbeat_registry": None,
     }
 
     try:
-        yield state
+        from acidwatch_api.broker.heartbeat import HeartbeatRegistry
+
+        registry = HeartbeatRegistry()
```
lars-petter-hauge: Is this the right place to put the heartbeat? I think I might have specified that it should be in the lifespan, but I was thinking lifespan of asyncio runner, not the specific function named _lifespan

Also, we are initiating the heartbeat in the database.py. I would much rather have both heartbeat listener and resultlistener and backend api started in the same location.

backend/src/acidwatch_api/routes/models.py
```diff
@@ -189,6 +183,21 @@ def get_models(
     return models
 
 
+@router.get("/models/status")
+def get_models_status(
```
lars-petter-hauge, line 165: instead of exposing the job ids per worker, we could attach the status on the result endpoints. That is, check if any workers are working on any of the simulations.

We don't want to leak implementation out of the backend of something called job ids, and have the caller side matching up what is working with something and checking if it is one of their simulations..

backend/tests/broker/integration/conftest.py
```diff
@@ -0,0 +1,340 @@
...
+the RabbitMQ implementation), but nothing outside this file/module should
+name it directly.
+
+Design under test (not yet implemented):
+
+    acidwatch_messaging.transport.RabbitMQTransport
+    acidwatch_messaging.queues
+    acidwatch_api.broker.heartbeat.HeartbeatRegistry
+    acidwatch_api.broker.listener.heartbeat_listener
+    acidwatch_api.broker.listener.result_listener
```
lars-petter-hauge: should rewrite this section. also this is wrong, as it is from when we did the tests before implementation.

backend/src/acidwatch_api/broker/listener.py
```diff
@@ -0,0 +1,149 @@
...
+    RESULTS_QUEUE,
+    Transport,
+    job_queue_name,
+)
+from acidwatch_models.datamodel import Conditions, Phase
+
+logger = logging.getLogger(__name__)
+
+
+async def restart_listener_on_failure(
```
lars-petter-hauge: I wonder if this might be too early addition, perhaps a bit unnecessary. Get some experience with it without a restart listener and see if we actually need it.

lars-petter-hauge: I ended up removing this part, and we can rather add it later if we deem it necessary

PR #811, opened 2026-08-10
Add SRK-VanLaar model support

Sviatose:
## Summary
Add a new SRK-VanLaar model adapter based on `neqsim` and remove the `solubilityccs` dependency from the backend dependency set.

## What Changed
- Added new adapter: `backend/src/acidwatch_api/models/srk_vanlaar.py`
- Registered `SRKVanLaarAdapter` in model exports and model routing
- Removed `solubilityccs` from backend dependencies
- Refreshed `uv.lock` after dependency updates
- Updated `neqsim` to the latest version used by the backend

## Behavior Notes
- The new adapter keeps the same overall output shape as the previous solubility flow (phase list + text panel)
- `HNO3` is now added correctly when `H2SO4` is also present
- All provided components are added to the VanLaar system input


lars-petter-hauge:
Awesome work!

So this means that the calculations done in solubilityccs is now integrated into jneqsim?

Happy to see that the model is now able to handle both acids present at the same time :)

Sviatose:
> Awesome work!
> 
> So this means that the calculations done in solubilityccs is now integrated into jneqsim?
> 
> Happy to see that the model is now able to handle both acids present at the same tim

> Awesome work!
> 
> So this means that the calculations done in solubilityccs is now integrated into jneqsim?
> 
> Happy to see that the model is now able to handle both acids present at the same time :)

Yes its now in jneqsim :)

lars-petter-hauge:
Did a quick test locally to verify it works, using default concentrations. Noticed there are variations (see how much H2O is added to acid phase), are those to be expected as we are now handling both acids at the same time? (although neither states that any HNO3 is added to the acidic phase)

Screenshot 2026-08-18 at 11 35 47

Screenshot 2026-08-18 at 11 35 51

Sviatose:
Hi,

Yes, it does look a bit odd. However, this is what the model predicts: when HNO3 levels increase dramatically to several thousand ppm, noticeable amounts appear in the acid phase. Currently, the model shows only about 0.02 wt% HNO3 in the formed acid in your example.

The high solubility of HNO3 was also indicated by IFE and confirmed by Rotvoll experiments. Perhaps we can include this as a discussion point at the next workshop. I think I also need to present you how the model works so we can discuss, but yeah, it currently shows extremely high solubility of HNO3. The question is then if we even need to focus on it or not

Sviatose:
image
at 10 bar and -29 (ship transport in gas phase) might is a bit worse with solubility of HNO3, so we see some amount 3% of HNO3 but still not a lot

lars-petter-hauge:
> it currently shows extremely high solubility of HNO3

From what I am reading at the results it shows the opposite? Default settings have almost no HNO3 in acidic phase. The curious thing is that much more H2O however is transferred to acidic phase.

Sviatose:
> > it currently shows extremely high solubility of HNO3
> 
> From what I am reading at the results it shows the opposite? Default settings don't have any HNO3 in acidic phase. The curious thing is that much more H2O however is transferred to acidic phase.

HNO3 extremely high solubility in CO2. Water has lower solubility in CO2 and H2SO4 is basically insoluble in CO2. Thats the current model results

lars-petter-hauge:
Alright. Just to be clear about my observations, completely disregarding HNO3, there is a clear difference of how much H2O is added to the acidic phase. Simplified example with only H2O and H2SO4:

Screenshot 2026-08-18 at 12 32 14

Screenshot 2026-08-18 at 12 32 29

Sviatose:
Yes, that is true. We currently lack sufficient experimental data for H₂SO₄ to accurately determine the amount of water migrating to the acidic phase. But I believe SolubilityCCS provides a bit more reliable estimate of H₂SO₄ solubility for now, especially at high concentrations.

lars-petter-hauge:
Alright, I am reading this as it is good to be merged :)

lars-petter-hauge (approved):

Issue #782, opened 2026-06-24
Refactor possibilities


lars-petter-hauge:
There are always things to do, but here are some I notice after the last days work:

 - The components are separated into simulation, grid-simulation and "the rest". We can easily split out for instance plots, and more.
 - We have plots and we have charts, pick one and stay with it
 - The GridPhaseChart has both a chart and a table. This is an easy split, but also, important to  check other similar components.
 - The Models page now does too much (after grid simulation was added), the component is like 200 lines long. Much of the content is duplicated depending on if it is grid mode or regular simulation. It would be very good if we could generalise over this.
 - Compare page has compare component for simulation inside, while gridsimulation is added to different page. Lets try to place these so that it makes sense. Also, they likely share some functionality.
- the helper function getCo2RichConcentrations is used throughout the codebase. I reckon that we are struggling a bit with the fact that we introduced phases, but for instance the input is always single phase (co2 feed). 
- We do in multiple places refer to the first of the simulation results. This hardcoded lookup looks like something that will eventually trip us.
- Several of the files in functions seems out of place, like Tables or GridSimulation (as most of the logic in the simulations are in the component. Perhaps its the other way round, the functions directory should have more logic related to simulations. Regardless, it's not consistent)

A simulation is just a subset of a simulation, lets see if we can incorporate that into the code.

Edit: also noticed that for the backend the routes folder is getting off. It contains a models.py and a gridsimulation.py. Both of them do simulations and gridsimulation.py uses things from models.py. Essentially, both naming of the files and the content of them is off. Requires a bit of restructure.

lars-petter-hauge:
knip results:

```
Unused files (4)
src/assets/syntheticResults.ts
src/components/Simulation/ConcResultTable.tsx
src/inject-env-template.js
vitest.setup.ts

Unused devDependencies (1)
@testing-library/user-event  package.json:46:10

Unused exports (14)
mergePhasesConcentrati…  function  …tion/MassBalanceError.tsx:50:17
AxisRange                          src/dto/GridSimulation.ts:4:14
getCo2RichPhase                    …/dto/SimulationResults.ts:39:14
EQUINOR_CHART_PALETTE              src/functions/Colors.ts:3:14
LAB_RESULT_COLOR_FAMIL…            src/functions/Colors.ts:11:14
filterInValidAndUndefi…  function  src/functions/Filtering.ts:13:17
filterAcidWithLowConce…  function  src/functions/Filtering.ts:31:17
[...]

PR #821, opened 2026-08-19
Refactor frontend

lars-petter-hauge:
worked on https://github.com/equinor/acidwatch/issues/782

Not completed, early stuff with copilot help. Likely requires some more minor things, and perhaps a couple commits removed (have not decided completely).

But the structure is improved at least (potentially we have too much going on for our web page, we would like it simple obviously. An improved structure might make it easier to take those decisions as well):

Screenshot 2026-08-19 at 15 34 01


lars-petter-hauge:
Still some things that could be improved (like gridphasetable still part of gridchart, needs another refactoring to get out). I will merge as it is now though, and do more later.

Removed some of the commits made earlier as I think they were just masquerading the complexity, instead of actually fixing it.

PR #837, opened 2026-09-04
change simulationgroup - simulation references database

lars-petter-hauge:
Changes how grid-simulation is structured in the database. Instead of saving the gridsimulation with an entry with a list of simulations, we save the gridsimulation id each child simulation (so... the correct way of doing it..).


backend/src/acidwatch_api/database.py
```diff
@@ -45,16 +45,26 @@ class Simulation(Base):
...
 class GridSimulation(Base):
     __tablename__ = "grid_simulations"
 
     owner_id: Mapped[UUID | None] = mapped_column(Uuid)
     axes: Mapped[list[dict]] = mapped_column(JSON)
-    simulation_ids: Mapped[list[str]] = mapped_column(JSON)
+
+    simulations: Mapped[list[Simulation]] = relationship(
+        back_populates="grid_simulation", order_by="Simulation.grid_position"
+    )
```
lars-petter-hauge, line 67: Main change in the PR. The simulation knows about gridsimulation, not the other way round. Correct way to handle many to one relationship..

backend/src/acidwatch_api/routes/_helpers.py
```diff
@@ -129,7 +129,30 @@ def query_chain_rows(
...
+        select(db.ModelInput, db.ModelResult)
+        .where(db.ModelInput.simulation_id.in_(simulation_ids))
+        .outerjoin(db.ModelResult)
+    )
+    grouped: dict[UUID, list[tuple[db.ModelInput, db.ModelResult | None]]] = (
+        defaultdict(list)
+    )
+    for model_input, result in session.execute(q).fetchall():
+        grouped[model_input.simulation_id].append((model_input, result))
+    return grouped
```
lars-petter-hauge: Single DB fetch for a gridsimulation, as opposed to 101 for a gridsimulation with 100 simulations..

backend/src/acidwatch_api/routes/grid_simulations.py
```diff
@@ -89,17 +90,16 @@ async def run_grid_simulation(
...
-        simulation_ids.append(str(simulation.id))
+        simulations.append(simulation)
 
         scheduled.append((point_concentrations, model_input_rows[0]))
 
     grid = db.GridSimulation(
         owner_id=UUID(user.id) if user else None,
         axes=[axis.model_dump() for axis in create.axes],
-        simulation_ids=simulation_ids,
+        simulations=simulations,
```
lars-petter-hauge, line 105: We let SQL alchemy handle the correct way of submitting entries. Since SQL Alchemy knows the dependencies this setup will ensure that in practice the GridSimulation entry enters first, and then the simulations. Technically it should be a single call to the db.

One minor issue is that if there is an error in saving a modelinput, the previous one would fail at the exact issue modelinput, while the new will fail with one large INSERT. I still think this is much nice though.

pinkwah: The way I see it, if we get a validation error from the database, we fucked up somewhere before. The database is there to pick up the worst problems.

On the other hand, if creating the grid failed but not the individual simulations, should those simulations still exist in the db? I reckon not and it's good that either everything is added or nothing.

backend/src/acidwatch_api/routes/_helpers.py
```diff
@@ -129,7 +129,30 @@ def query_chain_rows(
...
         .outerjoin(db.ModelResult)
     )
-    return [(row[0], row[1]) for row in session.execute(q).fetchall()]
+    return [
+        (model_input, model_result)
+        for model_input, model_result in session.execute(q).fetchall()
+    ]
+
+
+def query_chain_rows_by_simulation(
```
pinkwah: I don't understand the point of this function. The only usage of it is to fetch results for a single grid row. The way I initially read it, it's that we want to fetch simulation IDs from disparate grid simulations.

Anyway, since this is only used in the context of grid simulation, should the function not take a grid ID and go from there?

backend/alembic/versions/5fba0c16b695_restructure_grid_simulations.py
```diff
@@ -0,0 +1,75 @@
...
+    op.create_foreign_key(None, "simulations", "grid_simulations", ["grid_id"], ["id"])
+
+    # Backfill grid_id/grid_position from the old simulation_ids array,
+    # preserving the order each id appeared in.
+    op.execute("""
+        UPDATE simulations
+        SET grid_id = grids.id,
+            grid_position = members.position - 1
+        FROM grid_simulations AS grids,
+             jsonb_array_elements_text(grids.simulation_ids::jsonb)
```
pinkwah: I'm not entirely sure why we're converting simulation_ids to JSONB (`::jsonb`). If it's just `json` and not `jsonb`, then `json_array_elements_text(grid.simulation_ids)` ought work. Basically I'm trying to determine if this is a weird AI decision or if it's needed.

backend/src/acidwatch_api/routes/grid_simulations.py
lars-petter-hauge: The previous solution would do it the same way, as in either all is good or it is not added. It flushed after each entry, but the commit was in the end. Hence the transaction overall handled it the same way.

So I am not improving that part in any way. Would be nice if I did, but we did think about it before as well.. 😅

backend/alembic/versions/5fba0c16b695_restructure_grid_simulations.py
lars-petter-hauge: This could just be claude thingy. When I (clauded) a script to look up simulations and modelinputs etc, it was necessary to cast to json bytes. Can try to revert and check

backend/src/acidwatch_api/routes/_helpers.py
lars-petter-hauge: Yepp, I see that we are doing a call to get simulation ids first. Do not intend to fetch from different grid simulations. Will try to make it run on a single db call given a grid simulation id.

backend/alembic/versions/5fba0c16b695_restructure_grid_simulations.py
lars-petter-hauge: This was not needed. I will revert it.

I did check into it though:

> The json and jsonb data types accept almost identical sets of values as input. The major **practical difference is one of efficiency.** The json data type stores an exact copy of the input text, which processing functions must reparse on each execution; while **jsonb data** is stored in a decomposed binary format that makes it **slightly slower to input due to added conversion overhead, but significantly faster to process,** since no reparsing is needed. jsonb also supports indexing, which can be a significant advantage.
> 
> Because the json type stores an exact copy of the input text, it will preserve semantically-insignificant white space between tokens, as well as the order of keys within JSON objects. Also, if a JSON object within the value contains the same key more than once, all the key/value pairs are kept. (The processing functions consider the last value as the operative one.) By contrast, jsonb does not preserve white space, does not preserve the order of object keys, and does not keep duplicate object keys. If duplicate keys are specified in the input, only the last value is kept.
> 
> **In general, most applications should prefer to store JSON data as jsonb,** unless there are quite specialized needs, such as legacy assumptions about ordering of object keys.

https://www.postgresql.org/docs/current/datatype-json.html

Maybe we should actually consider jsonb then?

I reckon it might affect debugging, cause now we can see the input in plaintext making it significantly easier when looking in the database.

We have considered doing caching, and then we might do lookup on json content. When this occurs we will likely want to change, as jsonb supports equality checks (claude says json doesn't, I could only find that jsonb does from the manual)

> However, jsonb will preserve trailing fractional zeroes, as seen in this example, even though those are semantically insignificant for purposes such as equality checks.

**For this PR though, I will revert change and keep it as json.**

backend/src/acidwatch_api/routes/_helpers.py
lars-petter-hauge: you know what, the build_simulation_results also does another fetching from the database. That is, current approach would fetch twice per simulation. So a 100 simulation grid simulation would do 201 fetches to the database. We have a 1 second delay on frontend, so this quickly adds up.

Even though the postgres will handle this load, as it adds it to a queue, it could make the backend queue up and hence become slow (since we don't have scaling on backend). 

I will try to make modifications to reduce number of calls to database a lot (especially make it scale with number of simulations in a grid simulations)

lars-petter-hauge: I ended up renaming a bit as well, as we had implementation details (like chain_row) in function names rather than what it is (input_result pair). Also, did another refactoring on build_simulation_results, as it did too many things at once. Hopefully improved overall now

backend/src/acidwatch_api/routes/_helpers.py
```diff
@@ -164,23 +225,15 @@ def _phases_to_concentrations(phases: list[Phase]) -> dict[str, int | float]:
 
 
 def build_simulation_result(
-    session: Session,
-    simulation_id: UUID,
+    simulation: db.Simulation,
+    input_results: list[tuple[db.ModelInput, db.ModelResult | None]],
```
lars-petter-hauge, line 235: Now this function only makes the datastructures. It does not need a db session to build results.

PR #581, opened 2025-11-17
Facilitate adapter version

lars-petter-hauge:
solves: #488 

We don't know the version until the model has run (as it can change dynamically), hence we put it on the SimulationResult. Slight awkward that e.g. model_id and model_version are attached to two different pydantic objects (only mapped through Simulation_id), but that's perhaps something we need to live with.

PR #843, opened 2026-09-17
Changing ARCS and ARCS_exp descriptions 

badw:
changed the descriptions for ARCS and ARCS_exp. 

included `frontend/public/images` directory with `arcs_outer_loop.png` (not sure if this was the right place to put it, but it worked) 

changed ARCS and ARCS_Exp titles to ARCS 1.0.0 (Legacy) and ARCS 1.5.1 for more clarity, however ARCS and ARCS_exp are still labelled arcs and arcs_exp everywhere in the code. Started down that rabbit hole, regretted it, circled back. 

As far as I can tell, acid watch works as expected with no errors.


lars-petter-hauge:
Awesome :) Really like the updated description, we should probably improve even more how we do this :) 

(perhaps we should consider making it possible to build on documentation existing in the repository, so that we don't need duplicate docs).

Also, we do have the goal to have only a single arcs model.

Final comment, versions should be something acidwatch knows about specifically, not just in the description. Something like we did here: https://github.com/equinor/acidwatch/pull/581 (closed though)

